PR #1, opened 2019-10-24
Adding simple kokkos example

dialecticDolt:
Adding a really barebones cython wrapper for a c++ kokkos function.  
The current example is just a simple reduction but it shows interoperability with different devices.
Compilation scripts are set to assume 1 CPU and 4 GPU (Pascal) devices on node.


insertinterestingnamehere:
LGTM. Thanks!

PR #4, opened 2020-11-09
Working virtual execution environments

insertinterestingnamehere:


insertinterestingnamehere:
@arthurp whenever you get to it, do you mind looking at the merge conflict here? It's in your changes to the spawn interface, so I'm not 100% sure what the right version is.

PR #5, opened 2020-12-01
Start running threads after the whole threadpool got initialized

bozhiyou:


parla/task_runtime.py
```diff
@@ -638,6 +637,8 @@ def __init__(self, environments: Collection[TaskEnvironment], n_threads: int = N
         self._unassigned_resources = ResourcePool(multiplier=max_worker_queue_depth*1.0)
         self._worker_threads = [WorkerThread(self, i) for i in range(n_threads)]
         self._should_run = True
```
bozhiyou, line 639: _should_run is marked True during `super().__init__()` (of ControllableThread) so this remarking can also be removed?

arthurp: Good catch. Yes. Go ahead and do that in this PR if you like.

arthurp (approved):
This is simpler than I thought. Run the pytests. If they pass, LGTM.

bozhiyou:
Where to run the pytests? (Not to flush the Slack announcement so asked here)

If I barely run `pytest` in the root of Parla, this results in module and OS errors w/ or w/o changes of this pr. If I run `pytest` in tests, it hangs on test_tasks.py w/ or w/o the changes.

arthurp:
You should run pytest in the project root. And you will need to either install the library with `python setup.py install` or set PYTHONPATH to point to the project root, so it can find the modules. If you still have issues, then it's an issue that we need to address in general, since it will prevent everyone else from running pytests as well. Please post on slack and Ian may be able to help you.
Gotta focus on day job, I'll reappear late tonight or maybe Weds night.

bozhiyou:
No hurries.
`pytest` failed in my environment even in the master branch, so I have no idea if the changes hurt something. But these changes are tiny and easy to revert, so I'll merge this pr before further pushing updates to the branch (this branch is originally for auto data movement and I should have made a new branch for this pr).

parla/tasks.py
```diff
@@ -385,7 +385,7 @@ def spawn(taskid: Optional[TaskID] = None, dependencies = (), *,
     # TODO: Document tags argument
 
     if not taskid:
-        taskid = TaskID("global_" + str(len(_task_locals.global_tasks)), len(_task_locals.global_tasks))
```
bozhiyou, line 388: Line 46 hints id (2nd arg) to be Iterable[int] but here passed an int.

PR #20, opened 2021-03-30
Add initialize_thread to component instances.

arthurp:
This should initialize the thread for the component and is called
for every component on every thread.

@insertinterestingnamehere  needs to implement that function correctly for CuPy.

### DIFF
--- a/parla/cuda.py
+++ b/parla/cuda.py
@@ -167,6 +167,16 @@ class GPUComponentInstance(EnvironmentComponentInstance):
         self.stream.synchronize()
         return self.gpu.cupy_device.__exit__(exc_type, exc_val, exc_tb)
 
+    def initialize_thread(self) -> None:
+        for gpu in self.gpus:
+            # Trigger cblas/etc. initialization for this GPU in this thread.
+            device = gpu.cupy_device
+            with device:
+                device.cublas_handle
+                device.cusolver_handle
+                device.cusolver_sp_handle
+                device.cusparse_handle
+
 class GPUComponent(EnvironmentComponentDescriptor):
     """A single GPU CUDA component which configures the environment to use the specific GPU using a single
     non-blocking stream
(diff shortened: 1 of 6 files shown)


insertinterestingnamehere:
LGTM. I'll add the cupy init piece here.

insertinterestingnamehere:
When I try to run this I see TypeError: Can't instantiate abstract class GPUComponentInstance with abstract methods initialize_thread. Is this a quick fix on your end or do you want me to look into it further?

insertinterestingnamehere:
@nicelhc13 I think https://github.com/ut-parla/Parla.py/pull/20/commits/4f616d8a9411865cb75fa2dfb7b4b88bacb075b9 fixes the cupy context manager problem and is a better long-term solution. What do you think?

arthurp:
> When I try to run this I see TypeError: Can't instantiate abstract class GPUComponentInstance with abstract methods initialize_thread. Is this a quick fix on your end or do you want me to look into it further?

I didn't put `initialize_thread` in enough classes. It's on `MultiGPUComponentInstance` but it also needs to be on `GPUComponentInstance`. You can just do the same thing in that other class, but you will not need the loop.



> I think [4f616d8](https://github.com/ut-parla/Parla.py/commit/4f616d8a9411865cb75fa2dfb7b4b88bacb075b9) fixes the cupy context manager problem and is a better long-term solution. What do you think?

I think this is a good solution for the moment, but it may not be enough. I store `Device` instances a bunch of places. I assumed they were safe to shared freely. I also think it's rather non-intuitive that they cannot be shared, so I hope that gets fixed. I think it would be hard to maintain the invariant of always creating a new `Device` since it's so unnatural.

insertinterestingnamehere:
Yah, it's easier if they just fix it upstream (or even just let me fix it). Hochan tested that fix and already found that it's insufficient, so we're just going to go with his temporary fix for now until we can either get this taken care of upstream or do a more thorough fix in our codebase.

insertinterestingnamehere:
Rebased to duplicate the needed code. I also removed the device context manager partial fix here pending more work on that later.

parla/cuda.py
```diff
@@ -167,6 +167,16 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.stream.synchronize()
         return self.gpu.cupy_device.__exit__(exc_type, exc_val, exc_tb)
 
+    def initialize_thread(self) -> None:
+        for gpu in self.gpus:
+            # TODO(ian): Trigger cblas/etc. initialization for this GPU in this thread.
```
arthurp, line 172: Seems like this isn't a TODO any more. Delete it. Same below.

arthurp (commented):
This LGTM. Go ahead and merge it when you are ready.

parla/cuda.py
insertinterestingnamehere: Will do.

insertinterestingnamehere:
@dialecticDolt can you see if this improves the scaling situation for you? I'm seeing only a slight improvement from this.

insertinterestingnamehere:
Either way, this is an improvement, so merging.

PR #21, opened 2021-03-30
Blocked cyclic based multi-GPU blocked cholesky + clone_here bug fix

nicelhc13:
This PR contains two items:
1) blocked cyclic based multi-GPU blocked cholesky using row partitioning
2) `clone_here` fix; `cupy.as_array()` could not handle multiple CUDA stream well. This PR replace it with asynchronous copy specifying a target stream. `_GPUMemory` managed CUDA device object and multiple threads shared this global object without synchronization. This PR removed on-demand CUDA context change while `clone_here()` with an assumption that each task is only assigned a single GPU. Note that these are not the final fixes. We expect `cupy` will open their fixes soon and we will move the context object to thread local memory to avoid data race.

### DIFF
--- a/examples/blocked_cholesky_gpu.py
+++ b/examples/blocked_cholesky_gpu.py
@@ -3,7 +3,6 @@ A naive implementation of blocked Cholesky using Numba kernels on CPUs.
 """
 
 import numpy as np
-#import cupy
 from numba import jit, void, float64
 import math
 import time
@@ -64,13 +63,13 @@ def ltriang_solve(a, b):
         raise ValueError("Input array shapes are not compatible.")
     if a.shape[0] != a.shape[1]:
         raise ValueError("Array for back substitution is not square.")
-    #For the implementation here, just assume lower triangular.
+    # For the implementation here, just assume lower triangular.
     for i in range(a.shape[0]):
         b[i] /= a[i,i]
         b[i+1:] -= a[i+1:,i:i+1] * b[i:i+1]
     return b.T
 
-#comments would repack the data to column - major
+# comments would repack the data to column - major
 def cupy_trsm_wrapper(a, b):
     cublas_handle = device.get_cublas_handle()
     trsm = cublas.dtrsm
@@ -120,7 +119,7 @@ def cholesky_blocked_inplace(a):
     if a.shape[0] != a.shape[1]:
         raise ValueError("Non-square blocks are not supported.")
 
-    #Define task spaces
+    # Define task spaces
     gemm1 = TaskSpace("gemm1")        # Inter-block GEMM
     subcholesky = TaskSpace("subcholesky")  # Cholesky on block
     gemm2 = TaskSpace("gemm2")        # Inter-block GEMM
@@ -128,7 +127,7 @@ def cholesky_blocked_inplace(a):
 
     for j in range(a.shape[0]):
         for k in range(j):
-            #Inter - block GEMM
+            # Inter - block GEMM
             @spawn(gemm1[j, k], [solve[j, k]], placement=loc)
             def t1():
                 out = clone_here(a[j,j])  # Move data to the current device
@@ -136,7 +135,7 @@ def cholesky_blocked_inplace(a):
                 out = update(rhs, rhs, out)
                 copy(a[j,j], out)  # Move the result to the global array
 
-        #Cholesky on block
+        # Cholesky on block
         @spawn(subcholesky[j], [gemm1[j, 0:j]], placement=loc)
         def t2():
             dblock = clone_here(a[j, j])
@@ -145,7 +144,7 @@ def cholesky_blocked_inplace(a):
 
         for i in range(j+1, a.shape[0]):
             for k in range(j):
-                #Inter - block GEMM
+                # Inter - block GEMM
                 @spawn(gemm2[i, j, k], [solve[j, k], solve[i, k]], placement=loc)
                 def t3():
                     out = clone_here(a[i,j])  # Move data to the current device
@@ -154,7 +153,7 @@ def cholesky_blocked_inplace(a):
                     out = update(rhs1, rhs2, out)
                     copy(a[i,j], out)  # Move the result to the global array
 
-            #Triangular solve
+            # Triangular solve
             @spawn(solve[i, j], [gemm2[i, j, 0:j], subcholesky[j]], placement=loc)
             def t4():
                 factor = clone_here(a[j, j])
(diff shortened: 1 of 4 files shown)


parla/cuda.py
```diff
@@ -50,13 +52,28 @@ class _GPUMemory(Memory):
     def np(self):
         return _DeviceCUPy(self.device)
 
+    def asarray_sync(self, src):
```
insertinterestingnamehere, line 55: I think you mean for this to be asarray_async? sync refers to the cuda semantics that we're trying to avoid.

parla/cuda.py
```diff
@@ -50,13 +52,28 @@ class _GPUMemory(Memory):
...
-            if cupy.cuda.Device() != getattr(target, "device", None):
-                logger.debug("Moving data: %r => %r", getattr(target, "device", None), cupy.cuda.Device())
-                return cupy.asarray(target)
-            else:
-                return target
+        # TODO Several threads could share the same device object.
+        #      It causes data race and CUDA context is incorrectly set.
+        #      For now, this remove assumes that one device is always
+        #      assigned to one task.
+        #with self.device._device_context():
```
insertinterestingnamehere, line 66: Just pushed an alternate fix for the context manager issue to the other PR. LMK what you think. Otherwise, LGTM.

parla/cuda.py
nicelhc13: Yes my bad. I put synchronize() before right after the copy. That's why I put that name. I am going to fix this name.

parla/cuda.py
nicelhc13: Fixed!

parla/cuda.py
nicelhc13: Fixed

examples/blocked_cholesky_gpu.py
```diff
@@ -59,21 +64,19 @@ def ltriang_solve(a, b):
         raise ValueError("Input array shapes are not compatible.")
     if a.shape[0] != a.shape[1]:
         raise ValueError("Array for back substitution is not square.")
-    # For the implementation here, just assume lower triangular.
+    #For the implementation here, just assume lower triangular.
```
arthurp, line 67: I think the style guides for Python generally have a space after `#` in comments, so there is no reason to remove all said spaces. ;-) This is not important though. Just a nit.

examples/blocked_cholesky_multigpu.py
```diff
@@ -0,0 +1,265 @@
...
+  dev_id   = i % num_gpus
+  local_id = i // num_gpus
+  gpu_arrs[dev_id][local_id][j] = v
+
+def set_gpu_memory_from_cpu(a, num_gpus):
+  for j in range(a.shape[0]):
+    dev_id   = j % num_gpus 
+    local_id = j // num_gpus 
+    with cp.cuda.Device(dev_id):
+      gpu_arrs[dev_id][local_id] = cp.array(a[j], copy=True)
```
arthurp, line 190: These functions can be defined in a class along with `cholesky_blocked_inplace` so the class can capture the value of `num_gpu` and clean up this code a bit.

Also I think the mess here means that a multi-device array abstraction type would be useful. This simple approach may well not be general enough, but we clearly need a way to work with and directly index arrays that are actually multiple arrays on different devices. User code should never have to have junk like this.

examples/blocked_cholesky_multigpu.py
```diff
@@ -0,0 +1,265 @@
...
+    print(end - start, "seconds")
+    logger.debug("Calculate done..")
+
+    for i in range(len(gpu_arrs)):
+      logger.debug("Device ", i, " arrays are on ", gpu_arrs[i].device);
+
+    for d in range(len(gpu_arrs)):
+      logger.debug("Device:", d, " swap array..")
+      gpu_arrs[d] = cp.swapaxes(gpu_arrs[d], 2, 1)
+      logger.debug("Device:", d, " swap done..")
```
arthurp, line 243: Why not use `@spawn` and get some parallelism here. :-)

parla/array.py
```diff
@@ -5,6 +5,7 @@
 # FIXME: This load of numpy causes problems if numpy is multiloaded. So this breaks using VECs with parla tasks.
 #  Loading numpy locally works for some things, but not for the array._register_array_type call.
 import numpy as np
+import cupy as cp
```
arthurp, line 8: We should avoid this if at all possible. IIRC, importing cupy here prevents running the code at all without CUDA installed.

parla/cuda.py
arthurp: The code in this PR is actually wrong because this function can be called with a different device active to copy data to a remote device instead of to the current device. Let's go with Ian's solution if at all possible.

parla/cuda.py
```diff
@@ -50,13 +52,28 @@ class _GPUMemory(Memory):
...
+        if type(target) is numpy.ndarray:
```
arthurp, line 67: Use `isinstance` for type checks in Python.

parla/array.py
```diff
@@ -131,7 +132,6 @@ def clone_here(source, kind=None):
     else:
         raise TypeError("Array required, given value of type {}".format(type(source)))
 
-
```
arthurp, line 134: This isn't important since we are in a hurry, but please clean your diffs to remove random changes like this. In this case it actually makes the code match the style guides less well.

arthurp (approved):
I left some important and some less important comments. But overall, LGTM.

parla/cuda.py
arthurp: I see in https://github.com/ut-parla/Parla.py/pull/20#issuecomment-810648174 that Ian's fix is not enough. So do whatever hack you need but make sure to document that this is breaking the semantics of the function in a big way, so we can go back and fix it.

examples/blocked_cholesky_gpu.py
insertinterestingnamehere: This is correct. The spaces should be there.

examples/blocked_cholesky_multigpu.py
insertinterestingnamehere: Yah, we need to think about how to do this better. A multi-device array abstraction is a huge code cost, so we might be better served by working out some kind of hook into dask to leverage what they've already implemented in some way. Alternatively we could work on expanding our partitioning functionality. IDK. It's worth thinking about.

insertinterestingnamehere:
I just double-checked, and I think all the remaining comments here are non-blocking. Should I just go ahead and merge this?

arthurp:
> I just double-checked, and I think all the remaining comments here are non-blocking. Should I just go ahead and merge this?

Can you fix the following quickly before merging:

* https://github.com/ut-parla/Parla.py/pull/21#discussion_r604498388
* https://github.com/ut-parla/Parla.py/pull/21#discussion_r604499617
* https://github.com/ut-parla/Parla.py/pull/21#discussion_r604501673 (Adding TODOs to track the semantic problem)

I don't want to lose those issues because they can bite us later.

insertinterestingnamehere:
@nicelhc13 do you mind taking care of those?

nicelhc13:
Sure guys. I cannot work on this now since the filesystem of the TACC servers are not working now; I have no multi-GPU servers except them now so I cannot test it. I will merge this ASAP when the servers work.

examples/blocked_cholesky_multigpu.py
nicelhc13: It looks very good point. But for now, if you don't mind, let me just postpone this work to the another PR because it is excluded from time measurement and some people are waiting for this PR.

insertinterestingnamehere:
LGTM now.

insertinterestingnamehere:
Thank you!

PR #22, opened 2021-04-02
Automatic data movement based on partitioners w/ moving semantics

bozhiyou:
The programming paradigm in Parla is assumed to use partitioners. Partitioners with moving semantics generates logical devices as **views**, instead of copies, of the corresponding part of the original data. These logical devices will be wrapped so that any copy/clone operation is monitored and the placement is under control.

### DIFF
--- a/examples/blocked_cholesky_gpu.py
+++ b/examples/blocked_cholesky_gpu.py
@@ -8,13 +8,13 @@ import math
 import time
 
 from parla import Parla, get_all_devices
-from parla.array import copy, clone_here
 
 from parla.cuda import gpu
 from parla.cpu import cpu
 
 from parla.function_decorators import specialized
 from parla.tasks import *
+from parla.ldevice import LDeviceGridBlocked
 
 import cupy as cp
 from cupy.cuda import cublas
@@ -24,8 +24,8 @@ from cupy.linalg import _util
 from scipy import linalg
 import sys
 
-block_size = int(sys.argv[1])
-n = block_size*int(sys.argv[2])
+block_size = int(sys.argv[1]) if len(sys.argv) > 1 else 32*5
+n = block_size*int(sys.argv[2]) if len(sys.argv) > 2 else block_size*16
 
 loc = gpu
 
@@ -38,15 +38,15 @@ def cholesky(a):
     if a.shape[0] != a.shape[1]:
         raise ValueError("A square array is required.")
     for j in range(a.shape[0]):
-        a[j,j] = math.sqrt(a[j,j] - (a[j,:j] * a[j,:j]).sum())
+        a[j, j] = math.sqrt(a[j, j] - (a[j, :j] * a[j, :j]).sum())
         for i in range(j+1, a.shape[0]):
-            a[i,j] -= (a[i,:j] * a[j,:j]).sum()
-            a[i,j] /= a[j,j]
+            a[i, j] -= (a[i, :j] * a[j, :j]).sum()
+            a[i, j] /= a[j, j]
     return a
 
 
 @cholesky.variant(gpu)
-def choleksy_gpu(a):
+def cholesky_gpu(a):
     a = cp.linalg.cholesky(a)
     if cp.any(cp.isnan(a)):
       raise np.linalg.LinAlgError
@@ -65,8 +65,8 @@ def ltriang_solve(a, b):
         raise ValueError("Array for back substitution is not square.")
     # For the implementation here, just assume lower triangular.
     for i in range(a.shape[0]):
-        b[i] /= a[i,i]
-        b[i+1:] -= a[i+1:,i:i+1] * b[i:i+1]
+        b[i] /= a[i, i]
+        b[i+1:] -= a[i+1:, i:i+1] * b[i:i+1]
     return b.T
 
 # comments would repack the data to column - major
@@ -114,9 +114,10 @@ def cholesky_blocked_inplace(a):
     the block (row first, then column). The third and fourth index
     select the entry within the given block.
     """
-    if a.shape[0] * a.shape[2] != a.shape[1] * a.shape[3]:
+    # TODO (bozhi): these should be guaranteed by the partitioner
+    if len(a) * a[0][0].shape[0] != len(a[0]) * a[0][0].shape[1]:
         raise ValueError("A square matrix is required.")
-    if a.shape[0] != a.shape[1]:
+    if len(a) != len(a[0]):
         raise ValueError("Non-square blocks are not supported.")
 
     # Define task spaces
@@ -125,43 +126,43 @@ def cholesky_blocked_inplace(a):
     gemm2 = TaskSpace("gemm2")        # Inter-block GEMM
     solve = TaskSpace("solve")        # Triangular solve
 
-    for j in range(a.shape[0]):
+    for j in range(len(a)):
         for k in range(j):
             # Inter - block GEMM
             @spawn(gemm1[j, k], [solve[j, k]], placement=loc)
             def t1():
-                out = clone_here(a[j,j])  # Move data to the current device
-                rhs = clone_here(a[j,k])
+                out = a[j, j]
+                rhs = a[j, k]
                 out = update(rhs, rhs, out)
-                copy(a[j,j], out)  # Move the result to the global array
+                a[j, j] = out
 
         # Cholesky on block
         @spawn(subcholesky[j], [gemm1[j, 0:j]], placement=loc)
         def t2():
-            dblock = clone_here(a[j, j])
+            dblock = a[j, j]
             dblock = cholesky(dblock)
-            copy(a[j, j], dblock)
+            a[j, j] = dblock
 
-        for i in range(j+1, a.shape[0]):
+        for i in range(j+1, len(a)):
             for k in range(j):
                 # Inter - block GEMM
                 @spawn(gemm2[i, j, k], [solve[j, k], solve[i, k]], placement=loc)
                 def t3():
-                    out = clone_here(a[i,j])  # Move data to the current device
-                    rhs1 = clone_here(a[i,k])
-                    rhs2 = clone_here(a[j,k])
+                    out = a[i, j]
+                    rhs1 = a[i, k]
+                    rhs2 = a[j, k]
                     out = update(rhs1, rhs2, out)
-                    copy(a[i,j], out)  # Move the result to the global array
+                    a[i, j] = out
 
             # Triangular solve
             @spawn(solve[i, j], [gemm2[i, j, 0:j], subcholesky[j]], placement=loc)
             def t4():
-                factor = clone_here(a[j, j])
-                panel = clone_here(a[i, j])
+                factor = a[j, j]
+                panel = a[i, j]
                 out = ltriang_solve(factor, panel)
-                copy(a[i, j], out)
+                a[i, j] = out
 
-    return subcholesky[a.shape[0]-1]
+    return subcholesky[len(a)-1]
 
 def main():
     @spawn(placement=cpu)
@@ -175,7 +176,8 @@ def main():
 
         # Copy and layout input
         a1 = a.copy()
-        ap = a1.reshape(n // block_size, block_size, n // block_size, block_size).swapaxes(1,2)
+        mapper = LDeviceGridBlocked(n // block_size, n // block_size, devices=get_current_devices())
+        ap = mapper.partition_tensor(a1)
         start = time.perf_counter()
 
         # Call Parla Cholesky result and wait for completion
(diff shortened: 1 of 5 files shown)


examples/blocked_cholesky_gpu.py
```diff
@@ -175,7 +176,8 @@ async def test_blocked_cholesky():
 
         # Copy and layout input
         a1 = a.copy()
-        ap = a1.reshape(n // block_size, block_size, n // block_size, block_size).swapaxes(1,2)
+        mapper = LDeviceGridBlocked(n // block_size, n // block_size, devices=get_current_devices())
```
bozhiyou, line 179: This is the trick that creates views - partitions on the current device are created as views.

parla/ldevice.py
```diff
@@ -341,3 +342,58 @@ def wrapper(*args, memory, device):
...
+                elif isinstance(i, Iterable):
+                    for v in i:
+                        traverse(data[v], prefix + (v,), rest)
+                else:
+                    assert isinstance(i, int)
+                    traverse(data[i], prefix + (i,), rest)
+            else:
+                ret.append(clone_here(data) if is_array(data) else data)
+
+        traverse(self._latest_view, (), index)
```
bozhiyou, line 371: This traversing can be extremely inefficient and I assume it's the major overhead currently.

parla/ldevice.py
```diff
@@ -9,10 +9,12 @@
 from abc import abstractmethod, ABCMeta
 from functools import reduce
 from math import floor, ceil
-from typing import List, Tuple, Collection, Mapping
+from typing import List, Tuple, Iterable, Collection, Mapping
 from warnings import warn
 
 from parla.device import get_all_devices, Device, Memory, MemoryKind
+from parla.tasks import get_placement_for_any # TODO (bozhi): move get_placement_for_xxx to parla.device?
```
arthurp, line 16: I suspect `placement` will need it's own module. It is an interface and cuts across multiple parts of the the system. You can use `__parla_placement__` to avoid requiring the `placement` module to import a bunch of stuff.

parla/ldevice.py
```diff
@@ -341,3 +342,58 @@ def wrapper(*args, memory, device):
...
+                    for v in range(i.start or 0, i.stop, i.step or 1):
+                        traverse(data[v], prefix + (v,), rest)
+                elif isinstance(i, Iterable):
+                    for v in i:
+                        traverse(data[v], prefix + (v,), rest)
+                else:
+                    assert isinstance(i, int)
+                    traverse(data[i], prefix + (i,), rest)
+            else:
+                copy(data, value)
```
arthurp, line 394: This code is nearly identical to lines starting at 357. Lets create a function that takes a function to apply at the leaf calls. Here the function will be `lambda data: copy(data, value)`. Above the function will be `lambda data: ret.append(clone_here(data) if is_array(data) else data)`. Prefix is never used in this function. Remove it.

parla/ldevice.py
arthurp: I don't think this will be as much of an overhead as you would think. The traversal creates views at intermediate views not copies. Let's not try to optimize without profiling (which should be easy with `cProfile`). If this is a performance issue, then we can revisit this. Sadly optimizing this will be quite hard since this code uses python dynamic dispatch a lot. One option could be to catch the data views in some form to avoid having to rerun the traversal, but the cache might not be that much faster because of the complexity of the data that would need to be catched.

parla/ldevice.py
```diff
@@ -341,3 +342,58 @@ def wrapper(*args, memory, device):
         def wrapper(*args, memory, device):
             return memory(data(*args))
     return wrapper
+
+
+class Coordinator():
```
arthurp, line 347: This is a bad name. I'm not at all sure what's going on here. I think `Coordinator` is actually a partitioned array abstraction for arrays that are split across devices. And if so, let's call it `PartitionedNDArray`.

Also please document the semantics of the class and it's methods in doc strings. It's not clear exactly what each function is doing.

* What is the location of the the values returned by `__getitem__` and `__setitem__`?
* What happens if I assign into the result of `__getitem__` (e.g., `local_array = partitioned_array[x]; local_array[:] = ...;`)? Is this write local only? (it seems like it is.)

I don't care at the moment what the semantics are. Just please write down what you implemented.

parla/ldevice.py
```diff
@@ -126,8 +127,8 @@ def partition(self, data, memory_kind: MemoryKind = None):
         :return: A Python list of objects returned by `data` and copied to the appropriate device.
         """
```
arthurp, line 128: You have changed the semantics of these functions to return special objects that do lazy copies (they are not quite normal views, but that's fine). Update the doc strings to reflect the changes. You may well have to update the docs in many other functions in this file. You can search for "copied to the appropriate device" since that's the part the changed more than anything. 

I know we are really tight on time. I would be happy with adding a TODO to each doc string that will need updating saying `Docs need updating. The semantics have changed. See PartitionedNDArray.`

arthurp (approved):
I really do want changes. But I know how tight we are on time, so feel free to just drop in TODOs to respond all my comments. I just don't want to lose them.

Feel free to merge with just TODOs and fix it in another PR later.

parla/ldevice.py
bozhiyou: Are `placement` and `device` the same concept in Parla? We already have a `device` module and the `get_placement_for_any/set/value` interfaces ultimately use `__parla_placement__` for some types.

parla/ldevice.py
bozhiyou: Totally agreed this is a bad name 😛. I was actually trying to be forward-looking - this wrapper should also do some bookkeeping of where the parts have been and thus "coordinate"s the movement.

Documentation is important since I have assumptions on the programming paradigm. Will focus on clarifying what I'm/'ll be doing.

parla/ldevice.py
bozhiyou: > Let's not try to optimize without profiling (which should be easy with cProfile).
Agreed and will do. Thanks for the suggestions!

parla/ldevice.py
bozhiyou: You're totally right that I shamelessly stole the traverse from there 🤣. I was actually going to camp it out and mulling over a correct abstraction. Your approach does make sense. The prefix might be useful for future bookkeeping stuff though it's obviously not used here.

bozhiyou:
> I really like your approach. The array wrapper is really clean. I like how it hides the moves and does them lazily and stuff. Quite nice.
> 
> That said, I do want changes. But I know how tight we are on time, so feel free to just drop in TODOs to respond all my comments. I just don't want to lose them.
> 
> Feel free to merge with just TODOs and fix it in another PR later.

Thanks for the feedback. Before merging, I'm going to adapt all the examples we currently have to this change. This PR will still be merged to a non-master branch and after fixing the overhead we can eventually release it and deprecate `copy` and `clone_here` at user/application level.

parla/ldevice.py
arthurp: A placement is a collection of devices, interpreted as a disjunction (A or B or C). Not just a device. Or could go in device.py though. It's a pretty simple abstraction. Up to you.

parla/ldevice.py
arthurp: As a general rule classes should be named based on what abstraction they provide, not how they implement that abstraction. So the class is a partitioned nd array, the underlying coordination doesn't matter to a user of the class.

parla/ldevice.py
bozhiyou: Let me put these placement-related interfaces into the `device` module for now and maybe further rename the module to be the broader `placement`. After all, I think it anyways makes more sense than nesting them in `tasks`.

bozhiyou: Leave it an issue #25.

insertinterestingnamehere:
More of a high-level question here? How can/should we integrate this quickly into the existing apps? Which apps can benefit from this?

examples/blocked_cholesky_gpu.py
```diff
@@ -125,43 +126,43 @@ def cholesky_blocked_inplace(a):
...
-        for i in range(j+1, a.shape[0]):
+        for i in range(j+1, len(a)):
             for k in range(j):
                 # Inter - block GEMM
                 @spawn(gemm2[i, j, k], [solve[j, k], solve[i, k]], placement=loc)
                 def t3():
-                    out = clone_here(a[i,j])  # Move data to the current device
-                    rhs1 = clone_here(a[i,k])
-                    rhs2 = clone_here(a[j,k])
+                    out = a[i, j]
```
insertinterestingnamehere, line 151: Just to confirm that I've understood: are we using the technique of indexing blocks, then elements? I like that approach, but I can also see value in indices referring directly to the indices in the logical array that's being partitioned.

insertinterestingnamehere:
Broadly speaking, putting movement logic into a more intelligent indexable object is a fantastic idea.

examples/blocked_cholesky_gpu.py
```diff
@@ -175,7 +176,8 @@ async def test_blocked_cholesky():
 
         # Copy and layout input
         a1 = a.copy()
-        ap = a1.reshape(n // block_size, block_size, n // block_size, block_size).swapaxes(1,2)
+        mapper = LDeviceGridBlocked(n // block_size, n // block_size, placement=get_current_devices())
+        ap = mapper.partition_tensor(a1)
```
bozhiyou, line 180: > ... are we using the technique of indexing blocks, then elements?

This was originally used. What we do here is letting the dicing process pass through the partitioner - not arbitrary user reshaping - so that the runtime is aware of the partitioning scheme.

We assume that partitioned blocks are fixed size views of the corresponding part of the original array - users cannot change the partitioning scheme without making another partitioner. To give some background, we used to allow users to arbitrarily partition arrays (with `reshape` etc.) and assume arbitrary accesses including cross-block slicing (e.g. [0:2] for block1 and [2:4] for block2 but accessing [1:3]). We used to try hard to capture this partitioning and slicing behavior at syntax or bytecode level but things became very tricky.

examples/blocked_cholesky_gpu.py
bozhiyou: See comment [here](https://github.com/ut-parla/Parla.py/pull/22/files#r606687763).

bozhiyou:
> More of a high-level question here? How can/should we integrate this quickly into the existing apps? Which apps can benefit from this?

The modified `examples/blocked_cholesky_gpu.py` is an integration example (ignore white space and spelling fixes). I'm working on other current public examples before merging this PR. The app just needs to make sure one thing: any partitioning operation is done by a partitioner rather than `reshape` etc. Any app that requires `clone_here` and `copy` can benefit from this. The benefit includes 1) reduced LOC due to removed `clone_here` and `copy`, 2) separation of data movement from task logics (i.e. users don't have to care about data movement), and 3) possible performance gain given that user-level movement is not cautious (e.g. task A and B are both on the same device and accessing the same data; A's writeback and B's clone_here are redundant).

parla/utils.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import TypeVar, Tuple, Union, Iterable, Callable
+
+__all__ = ["traverse",]
+
+
+# TODO (bozhi): another parla module for typing?
+T = TypeVar('T') # (PEP 484) The argument to TypeVar() must be a string equal to the variable name to which it is assigned.
+
+
+def traverse(prefix: T, index: Tuple[Union[slice, Iterable[int], int]], step: Callable[[T, int], T], stop: Callable[[T], None]):
```
arthurp, line 15: `traverse` is not specific enough of a name. This is a pretty specific operation related to in nd indicies. The function name should reflect that.

bozhiyou: It's a legacy name and I noted this as a todo in the docstring below (line 24). I will consider a better name when updating the docs.

insertinterestingnamehere:
Is there anything left to do on this one? I'd like to merge soon to let @sestephens73 exercise this interface in his QR app.

bozhiyou:
I'm profiling this for performance overhead. Should be done today and I'll adapt the QR app with this.

insertinterestingnamehere:
Sounds great. Thank you!

insertinterestingnamehere:
Also worth mentioning: the parla core runtime isn't especially well-optimized yet. So long as the performance of this is not terrible, it'll be fine to merge. We want to optimize the core runtime, but for now we're just relying on the tasks being big enough to amortize the scheduling costs.

bozhiyou:
Interesting enough, the "overhead" comes from `logger.debug`! I put a lot of them and even though the logging level mutes them, they still take significant time! removing them makes the performance match that of manual movement.

bozhiyou:
@insertinterestingnamehere ready to merge.

PR #27, opened 2021-04-04
Added number of iterations for tests on the multi-GPU Cholesky

nicelhc13:


examples/blocked_cholesky_multigpu.py
```diff
@@ -197,68 +205,76 @@ async def test_blocked_cholesky():
...
+      a = np.random.rand(n, n)
+      a = a @ a.T
+      logger.debug("Random number generate done..")
+
+      logger.debug("Copy a to a1..")
+      # Copy and layout input
+      a1 = a.copy()
+      logger.debug("Copying done..")
+      logger.debug("Shaping starts..")
+      ap = a1.reshape(n // block_size, block_size, n // block_size, block_size).swapaxes(1,2)
```
dialecticDolt, line 232: This is just a comment, otherwise it lgtm

iirc  this takes a lot of time to setup the matrix but I'm not sure what % of total runtime it is. 

A quality of life improvement for faster timing runs might be setting it up once at the beginning and creating copies of the data instead. (To avoid losing the GPU memory due to the extra copy you could store them in host and copy back at the start of a loop)

dialecticDolt (approved):

examples/blocked_cholesky_multigpu.py
nicelhc13: Good point. I will fix this :-)

PR #32, opened 2021-04-05
Create a new stream for each task.

### LINKED ISSUE #16: Tasks should run on their own streams
Title says it all. For some reason we didn't set it up like this. IIUC, this may limit parallelism when we co-locate kernels on a device.

arthurp:
Fixes #16

### DIFF
--- a/parla/cuda.py
+++ b/parla/cuda.py
@@ -97,7 +97,6 @@ class _GPUDevice(Device):
             yield
 
     @property
-    @lru_cache(None)
     def cupy_device(self):
         return cupy.cuda.Device(self.index)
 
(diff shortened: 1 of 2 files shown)


arthurp:
@insertinterestingnamehere Take a look. I think I solved this issue. And maybe some of the device object issues too. I only ran the unit tests though. Maybe some benchmark testing would be a good idea before merging.

parla/cuda.py
```diff
@@ -163,27 +162,68 @@ def get_array_module(self, a):
...
+    def push_device(self, dev):
+        self._device_stack.append(dev)
+
+    def pop_device(self) -> cupy.cuda.Device:
+        return self._device_stack.pop()
+
+    @property
+    def stream(self):
+        if self._stream_stack:
+            return self._stream_stack[-1]
```
insertinterestingnamehere, line 189: Just to confirm that I'm understanding right, the only reason this won't cause unintentional dependencies between GPU kernel calls is that whichever task most recently used a given stream already did stream.synchronize before pushing the stream back onto the stack?

parla/cuda.py
```diff
@@ -163,27 +162,68 @@ def get_array_module(self, a):
...
+        stream.__enter__()
+        self._stack.push_stream(stream)
+        self._stack.push_device(dev)
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.stream.__exit__(exc_type, exc_val, exc_tb)
-        self.stream.synchronize()
-        return self.gpu.cupy_device.__exit__(exc_type, exc_val, exc_tb)
+        stream = self._stack.pop_stream()
```
insertinterestingnamehere, line 221: I think there's an ordering issue here. Shouldn't we synchronize the stream, then exit it, then pop the stream, then pop the device.

insertinterestingnamehere (commented):
Mostly LGTM, modulo the synchronization issues mentioned in the comments. We should probably merge and then ask others to check for performance changes in some of the apps. I'll try it out really quick for some simple stuff though.

parla/cuda.py
arthurp: This property is actually never used. I meant to remove it.

Also streams are never reused. They are created, entered, pushed, popped, exited, then discarded. They are never shared.

insertinterestingnamehere: Excellent. Supposedly stream creation overhead is extremely small, so I'd much rather just run with a new one.

arthurp: It turns out that I need this function to implement the ordering you mention, but the semantics I stated are still true.

Issue #18, opened 2021-03-24
clone_here May Swap Current Device


insertinterestingnamehere:
Currently clone_here may switch out the current device unintentionally. An older version of the matmul  This is NOT desired behavior and needs fixing.

insertinterestingnamehere:
Note: a temporary workaround is to use cupy.asarray instead when the task involves copying from somewhere else to the current GPU. That workaround should suffice for many of the existing examples, however clone_here is supposed to work even when we're copying data from a GPU to the CPU or getting a view of data that's already located on the CPU.

insertinterestingnamehere:
On second thought, I'm not tagging this one for the paper deadline. We can get by without it since all the example apps can just use cupy.asarray or cupy.asnumpy to get what they need.

insertinterestingnamehere:
Note: we've isolated this one and reported it upstream as https://github.com/cupy/cupy/issues/5006.

insertinterestingnamehere:
A partial fix was merged in https://github.com/ut-parla/Parla.py/pull/21.

insertinterestingnamehere:
Hoping to get this fixed upstream in https://github.com/cupy/cupy/pull/5083.

insertinterestingnamehere:
Fixed upstream. Hopefully available in the 9.1 release in about a month.

insertinterestingnamehere:
cupy 9.1.0 is out. The fix for this now is to just update cupy.

PR #33, opened 2021-04-05
Fix data movement calls for partitioners

insertinterestingnamehere:
https://github.com/ut-parla/Parla.py/pull/21 included a partial fix for https://github.com/ut-parla/Parla.py/issues/18, but it unintentionally broke our partitioner objects. This fixes that. The underlying issue upstream is https://github.com/cupy/cupy/issues/5006.

### DIFF
--- a/parla/cuda.py
+++ b/parla/cuda.py
@@ -66,16 +66,17 @@ class _GPUMemory(Memory):
         # FIXME This code breaks the semantics since a different device
         #       could copy data on the current device to a remote device.
         #with self.device._device_context():
-        if isinstance(target, numpy.ndarray):
-            logger.debug("Moving data: CPU => %r", cupy.cuda.Device())
-            return cupy.asarray(target)
-        elif type(target) is cupy.ndarray and \
-             cupy.cuda.Device() != getattr(target, "device", None):
-            logger.debug("Moving data: %r => %r",
-                         getattr(target, "device", None), cupy.cuda.Device())
-            return self.asarray_async(target)
-        else:
-            return target
+        with cupy.cuda.Device(self.device.index):
+            if isinstance(target, numpy.ndarray):
+                logger.debug("Moving data: CPU => %r", cupy.cuda.Device())
+                return cupy.asarray(target)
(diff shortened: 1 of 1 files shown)


parla/cuda.py
```diff
@@ -66,16 +66,17 @@ def __call__(self, target):
...
-            logger.debug("Moving data: CPU => %r", cupy.cuda.Device())
-            return cupy.asarray(target)
-        elif type(target) is cupy.ndarray and \
-             cupy.cuda.Device() != getattr(target, "device", None):
-            logger.debug("Moving data: %r => %r",
-                         getattr(target, "device", None), cupy.cuda.Device())
-            return self.asarray_async(target)
-        else:
-            return target
+        with cupy.cuda.Device(self.device.cupy_device.id):
```
arthurp, line 69: Tou can write: `self.device.index`. This avoids the step through `cupy_device`.

arthurp (approved):
LGTM.

parla/cuda.py
```diff
@@ -66,16 +66,17 @@ def __call__(self, target):
...
+            if isinstance(target, numpy.ndarray):
+                logger.debug("Moving data: CPU => %r", cupy.cuda.Device())
+                return cupy.asarray(target)
+            elif type(target) is cupy.ndarray and \
```
nicelhc13, line 73: Can we use `isinstance(target, cupy.ndarray)`?

nicelhc13:
I was wondering if you have tested this version with an application like Blocked Cholesky. 
If it was fine, LGTM.

parla/cuda.py
insertinterestingnamehere: Sure. I'll change that. I was just indenting what was there before, but switching it over is fine.

parla/cuda.py
insertinterestingnamehere: Done. Thanks!

nicelhc13 (approved):
LGTM!

insertinterestingnamehere:
It appears to work fine on Cholesky. Merging.

Issue #19, opened 2021-03-24
Scheduler Awareness of Persistent Data


insertinterestingnamehere:
The scheduler needs to be aware of persistent data on each device, otherwise our occupancy model can't account for tasks that need to create some data and then leave it there or for the case where some persistent data exists on each device throughout an entire application. The easiest short-term fix is to just use a context manager to explicitly reserve space. The long-term fix is probably an array wrapper class, but that's significantly more work, so we can't do that for the current paper deadline.

insertinterestingnamehere:
There's potentially more to do here for user convenience, but we at least have an interface for this now, so closing.

PR #40, opened 2021-04-07
Initial working version of persistent data reservation.

insertinterestingnamehere:
Control flow is still a bit messy and tests need to be WAY more thorough, but this should work. @nicelhc13, I think Cholesky can use the same idiom as I show in matmul, but haven't actually confirmed. Try it out and let me know how that goes.

### DIFF
--- a/examples/matmul.py
+++ b/examples/matmul.py
@@ -63,8 +63,7 @@ def main():
                         @spawn(matmul[i, j], placement = c_block, memory = memsize)
                         def matmul_task():
                             old_device = cp.cuda.Device()
-                            #b_block_local = clone_here(b_block)
-                            b_block_local = cp.asarray(b_block, order = 'F')
+                            b_block_local = clone_here(b_block)
                             # cupy doesn't support the out argument for matmul yet so we have to copy.
                             # cp.matmul(a_block, b_block_local.T, out = c_block)
                             #print(i, j, old_device, cp.cuda.Device(), c_block.device, a_block.device, b_block_local.device, b_block.device)
--- a/examples/matmul.py
+++ b/examples/matmul.py
@@ -14,7 +14,7 @@ from parla.cpu import cpu
 from parla.cuda import gpu
 from parla.function_decorators import specialized
 from parla.ldevice import LDeviceSequenceBlocked
-from parla.tasks import spawn, TaskSpace, CompletedTaskSpace
+from parla.tasks import spawn, TaskSpace, CompletedTaskSpace, reserve_persistent_memory
 
 def main():
     ngpus = int(sys.argv[1])
@@ -44,32 +44,35 @@ def main():
             c_part.append(cp.empty((c_dim, n), np.float32, order = 'F'))
 
     previous = None
-    for repetition in range(repetitions):
-        # Now compute a @ b.T and write the output to c
-        deps = [previous] if previous is not None else []
-        @spawn(placement = cpu, dependencies = deps)
-        async def run_matmul():
-            print("start")
-            start = time.perf_counter()
-            matmul = TaskSpace("matmul")
-            for i in range(blocks):
-                for j in range(blocks):
-                    a_block = a_part[i]
-                    b_block = b_part[j]
-                    c_block = c_part[i][:, j * block_size : (j + 1) * block_size]
-                    @spawn(matmul[i, j], placement = c_block)
-                    def matmul_task():
-                        old_device = cp.cuda.Device()
-                        #b_block_local = clone_here(b_block)
-                        b_block_local = cp.asarray(b_block, order = 'F')
-                        # cupy doesn't support the out argument for matmul yet so we have to copy.
-                        # cp.matmul(a_block, b_block_local.T, out = c_block)
-                        #print(i, j, old_device, cp.cuda.Device(), c_block.device, a_block.device, b_block_local.device, b_block.device)
-                        c_block[:] = a_block @ b_block_local.T
-            await matmul
-            stop = time.perf_counter()
-            print(stop - start)
-        previous = run_matmul
+    with reserve_persistent_memory([a_part, b_part, c_part]):
+        for repetition in range(repetitions):
+            # Now compute a @ b.T and write the output to c
+            deps = [previous] if previous is not None else []
+            @spawn(placement = cpu, dependencies = deps)
+            async def run_matmul():
+                start = time.perf_counter()
+                matmul = TaskSpace("matmul")
+                for i in range(blocks):
+                    for j in range(blocks):
+                        a_block = a_part[i]
+                        b_block = b_part[j]
+                        c_block = c_part[i][:, j * block_size : (j + 1) * block_size]
+                        memsize = c_block.nbytes
+                        if i != j:
+                            memsize += b_block.nbytes
+                        @spawn(matmul[i, j], placement = c_block, memory = memsize)
+                        def matmul_task():
+                            old_device = cp.cuda.Device()
+                            #b_block_local = clone_here(b_block)
+                            b_block_local = cp.asarray(b_block, order = 'F')
+                            # cupy doesn't support the out argument for matmul yet so we have to copy.
+                            # cp.matmul(a_block, b_block_local.T, out = c_block)
+                            #print(i, j, old_device, cp.cuda.Device(), c_block.device, a_block.device, b_block_local.device, b_block.device)
+                            c_block[:] = a_block @ b_block_local.T
+                await matmul
+                stop = time.perf_counter()
+                print(stop - start)
+            previous = run_matmul
 
 if __name__ == "__main__":
     with Parla():
--- a/parla/tasks.py
+++ b/parla/tasks.py
@@ -448,7 +448,7 @@ def get_current_devices() -> List[Device]:
 
 
 @contextmanager
-def _reserve_persistent(memsize, device):
+def _reserve_persistent_memory(memsize, device):
     resource_pool = get_scheduler_context().scheduler._available_resources
     resource_pool.allocate_resources(device, {'memory' : memsize}, blocking = True)
     try:
@@ -456,7 +456,7 @@ def _reserve_persistent(memsize, device):
     finally:
         resource_pool.deallocate_resources(device, {'memory' : memsize})
 
-
+# TODO: Move this to parla.device and import it from there. It's generally useful.
 def _get_parla_device(device):
     if isinstance(device, Device):
         return device
@@ -486,6 +486,16 @@ def reserve_persistent_memory(amount, device = None):
       be supplied if any element of the list is not an array. This may be a list of \
       devices if amount is a list of array objects.
     """
+    # TODO: This function should be split up into simpler subunits.
+    # How exactly that should be done isn't settled yet, but there's
+    # some discussion on this at
+    # https://github.com/ut-parla/Parla.py/pull/40#discussion_r608857593
+    # https://github.com/ut-parla/Parla.py/pull/40#discussion_r608853345
+    # TODO: reduce nesting by separating out the try/except idioms for
+    # checking if something supports the buffer protocol and checking
+    # whether or not something is iterable into separate functions.
+    # TODO: Generalize the naming/interface here to allow reserving
+    # resources other than memory.
     from . import cpu
     if isinstance(amount, int):
         memsize = amount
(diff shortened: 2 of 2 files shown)


insertinterestingnamehere:
General note: really the only way this is currently going to work is if you run it inside the parla context manager before launching a bunch of GPU tasks. If you use this outside the context manager it'll error out and if you use it inside an app that's already running a bunch of tasks that reserve gpu memory it may deadlock.

insertinterestingnamehere:
Fixes https://github.com/ut-parla/Parla.py/issues/19

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+    try:
+        import cupy
+    except ImportError:
+        pass
+    else:
+        if isinstance(device, cupy.cuda.Device):
+            from .cuda import gpu
+            index = device.id
+            return gpu(index)
+    raise ValueError("Don't know how to convert object of type {} to a parla device object.".format(type(device)))
```
arthurp, line 472: This is OK for now. But please add a TODO that this needs to be made generic and moved to `device.py`. I suspect the implementation needs to be that `Architecture`s provide a method that attempts to convert an object to a Parla device, then a module level function tries all the `Architecture`s to find one that can convert the object, or fails. This could then be used anywhere a device is expected to provide Python style typing flexibility for device objects. All without having any reference to CUDA in the Parla core.

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+
+
+@contextmanager
+def reserve_persistent_memory(amount, device = None):
```
arthurp, line 476: OK for now, but add TODO. This should not be specific to memory, it should be possible to reserve any resource, so this should be called `reserve` or `reserve_resources` and should take keyword arguments for all resources (maybe via `**`).

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+    if isinstance(amount, int):
+        memsize = amount
+    elif hasattr(amount, '__cuda_array_interface__'):
+        import cupy
+        # cupy has the info we need and views are cheap,
+        # so get a view then read off the size/device info.
+        view = cupy.array(amount)
+        memsize = view.nbytes
+        if device is None:
+            device = view.device
```
arthurp, line 499: I think you can just use `storage_size` here. It's in `parla.array` I think.

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+                            arr.__cuda_array_interface__
+                        except AttributeError as exc:
+                            inner_must_be_iterable = True
+                        else:
+                            stack.enter_context(reserve_persistent_memory(arr))
+                        if inner_must_be_iterable:
+                            try:
+                                iter(arr)
+                            except TypeError as exc:
+                                raise ValueError("Implicit location specification only supported for GPU arrays.") from exc
```
arthurp, line 536: Why? There is `parla.array.get_memory(a).device`, which should do what you need without any of this.

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+            try:
+                device = _get_parla_device(device)
+            except ValueError:
+                device_must_be_iterable = True
+            if device_must_be_iterable:
+                with ExitStack() as stack:
+                    for arr, dev in zip(amount, device):
+                        stack.enter_context(reserve_persistent_memory(arr, dev))
+                    yield
+                    return
```
arthurp, line 551: I do NOT want to support this. It's confusing to have an implicit zip operation. If you want to have a variant of `reserve_persistent_memory` that accepts a dict or a collection of pairs that would be fine with me. Though I think it should be a different function. Too many ways to call a single function makes type errors impossible to debug. Numpy falls into this trap a lot, let's not follow them.

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+    :param amount: The number of bytes reserved in the scheduler from tasks for persitent data. \
+      This exists, not as any kind of enforced limit on allocation, but rather to let the scheduler \
+      have an accurate measure of memory occupancy on the GPU beyond just memory that's used \
+      only during a task's execution. It can be specified as an integer representing the nubmer of \
+      bytes, an ndarray (cupy or numpy), or a list of ndarrays.
+    :param device: The device object where memory is to be reserved. \
+      This must be supplied if amount is an integer \
+      and may be supplied for an array. In the case of a list or other iterable it must \
+      be supplied if any element of the list is not an array. This may be a list of \
+      devices if amount is a list of array objects.
```
arthurp, line 487: I think this is too flexible and has too many special cases. Can we split this into three functions:

* `reserve_data` takes `*args` where each is a data object (e.g., array) and reserves the size of each on it's device.
* `reserve_persistent_memory` takes a device and an amount and reserve it.
* `reserve_persistent_memories` takes a collection of pairs `(device, amount)` or dict and reserves every pair as above. (I would be OK with merging this one with the previous variant.)

As above, the memory functions would eventually be generalized into a function that can reserve any resource.

I just don't want Parla to end up line numpy with 6 different ways to call every function and different restrictions on each way to call it. It's hard to document and confusing, since functions don't have simple and concise semantics.

examples/matmul.py
```diff
@@ -44,32 +44,35 @@ def main():
...
+                matmul = TaskSpace("matmul")
+                for i in range(blocks):
+                    for j in range(blocks):
+                        a_block = a_part[i]
+                        b_block = b_part[j]
+                        c_block = c_part[i][:, j * block_size : (j + 1) * block_size]
+                        memsize = c_block.nbytes
+                        if i != j:
+                            memsize += b_block.nbytes
+                        @spawn(matmul[i, j], placement = c_block, memory = memsize)
```
arthurp, line 63: We should make `storage_size` remove duplicate or overlapping arrays from it's argument list, so this can just be `memory = storage_size(c_block, b_block)`.

arthurp (approved):
Please add todos for my comments, but then go ahead and merge due to the deadline. This should work, it's just a dangerous API design in my opinion and spills CUDA details into the core a lot, so it does need a good bit of work to make this acceptable in the long term.

parla/tasks.py
insertinterestingnamehere: Good point.

parla/tasks.py
insertinterestingnamehere: Sounds good.

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+    """
+    from . import cpu
+    if isinstance(amount, int):
+        memsize = amount
+    elif hasattr(amount, '__cuda_array_interface__'):
+        import cupy
+        # cupy has the info we need and views are cheap,
+        # so get a view then read off the size/device info.
+        view = cupy.array(amount)
```
nicelhc13, line 496: `cupy.array()` always copies the `amount` object. Is it expected? Otherwise, we can pass `copy=False` parameter.

insertinterestingnamehere: It always returns a new array object, it doesn't necessarily copy the underlying data. `copy=False` is still a good idea here though. All I'm doing is getting it to handle the case of some object that exposes `__cuda_array_interface__` but isn't actually a cupy array.

parla/tasks.py
insertinterestingnamehere: IMHO the ability to expand this operation automatically over lists of things is a major value add right now since the cases where we actually use this all involve times when we're storing lists of per-device blocks. The zip seemed like the most natural way to extend the device argument to match that use-case. We can get rid of the zip, but recursively expanding over lists doesn't fit in the list of functions you've provided below unless we generalize reserve_data a bit.

parla/tasks.py
insertinterestingnamehere: While I understand that criticism I don't totally agree with it. I personally find the mental load of having tons of functions with extremely similar use-cases marginally worse than having different ways to call a single function that still, conceptually, represents one kind of operation. My main criticism of the APIs upstream is when functions cease to even represent the operation you'd expect them to represent (e.g. how numpy.transpose is the designated interface for arbitrarily permuting axes). Either way, for now I'll just make a TODO note that references this discussion and move on.

examples/matmul.py
insertinterestingnamehere: I agree with that suggestion. In this case it wouldn't help though since the issue is that if b_block is already on the current device space was reserved for it elsewhere. Idk what a good fix for that would be. We could track reserved arrays somewhere, but I'm hesitant to add a global buffer pool like that.

parla/tasks.py
```diff
@@ -447,6 +447,124 @@ def get_current_devices() -> List[Device]:
...
+                device = cpu(0)
+        # Not a cpu array, so try handling amount as
+        # an iterable of things that each need to be processed.
+        if amount_must_be_iterable:
+            try:
+                iter(amount)
+            except TypeError as exc:
+                raise ValueError("Persistent memory spec is not an integer, array, or iterable object") from exc
+            if device is None:
+                with ExitStack() as stack:
```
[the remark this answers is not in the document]
nicelhc13: I really like the way you guys write python. It is like an art.

parla/tasks.py
insertinterestingnamehere: Great point. I wasn't aware that interface existed.

parla/tasks.py
insertinterestingnamehere: Good point. I wasn't aware of that function.

parla/tasks.py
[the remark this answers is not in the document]
insertinterestingnamehere: Thanks! I still want to come back to this and factor out some of the try/except blocks into utility functions to cut down the nesting, but I was really happy to find that someone had put together an idiom for entering multiple context managers.

parla/tasks.py
insertinterestingnamehere: Actually, no you're totally right on this one. The object will have its data copied too. `copy=False` fixes it.

insertinterestingnamehere:
Thanks for looking through this! I'll mark todos for most of the comments here then merge.

parla/tasks.py
arthurp: I'm OK with recursively flattening the arguments to `reserve_data`. The zip is the confusing part to me. I would be a little happier with cartesian product and relatively comfortable with implicitly mapping over *one* argument as long as the other is a single value.

parla/tasks.py
arthurp: My argument is that these different functions *don't* implement the same operation. But we will debate that later and in the end I'm fading out of the project, so my design opinions are less and less relaxant.

examples/matmul.py
arthurp: Hmm. OK. I think this is just another reason that we really need a PTask-like buffer tracking system, so that we can handle all this as part of the cache protocol.

parla/tasks.py
```diff
@@ -486,14 +486,27 @@ def reserve_persistent_memory(amount, device = None):
...
         memsize = amount
     elif hasattr(amount, '__cuda_array_interface__'):
         import cupy
         # cupy has the info we need and views are cheap,
         # so get a view then read off the size/device info.
-        view = cupy.array(amount)
+        view = cupy.array(amount, copy = False)
+        # TODO: there's actually a storage_size helper routine in
+        # parla.array. We should use it here and move the logic
+        # about buffer protocols there.
```
arthurp, line 509: The logic about buffer protocols goes in the buffer type implementations defined in `parla.cpu` and `parla.cuda`. `parla.array` has no concrete knowledge of how buffer types work.

insertinterestingnamehere: How do we want to consistently get the size of the buffers? Most things support itemsize/size/nbytes, but numba's DeviceNDArrays do not (https://numba.pydata.org/numba-doc/dev/cuda/memory.html#device-arrays). That's just one such example. There are lots of projects out there that use pep 3118/\_\_array_interface\_\_/\_\_cuda_array_interface\_\_ and our safest bet is to get memory use information through those kinds of protocols.

insertinterestingnamehere: I guess probably the takeaway is that this needs some more thought. Either way what we have is good enough for the time being.

parla/tasks.py
insertinterestingnamehere: Yah, I'm working toward moving on as well, so this decision can be left to others.

parla/tasks.py
insertinterestingnamehere: Yah, getting rid of the zip option is totally fine with me.

PR #41, opened 2021-04-07
Fix our asarray workaround.

insertinterestingnamehere:
We've been seeing excessive memory use and unintended transposes happening in our asarray workaround. This fixes those.

### DIFF
--- a/parla/cuda.py
+++ b/parla/cuda.py
@@ -53,10 +53,13 @@ class _GPUMemory(Memory):
         return _DeviceCUPy(self.device)
 
     def asarray_async(self, src):
-      dst = cupy.ndarray(src.shape)
-      dst.data.copy_from_device_async(src.data, src.nbytes,
-                                      cupy.cuda.get_current_stream())
-      return dst
+        if isinstance(src, cupy.ndarray) and src.device.id == self.device.index:
+            return src
+        if not (src.flags['C_CONTIGUOUS'] or src.flags['F_CONTIGUOUS']):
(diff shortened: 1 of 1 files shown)


parla/cuda.py
```diff
@@ -53,10 +53,13 @@ def np(self):
...
-      dst = cupy.ndarray(src.shape)
-      dst.data.copy_from_device_async(src.data, src.nbytes,
-                                      cupy.cuda.get_current_stream())
-      return dst
+        if isinstance(src, cupy.ndarray) and src.device.id == self.device.index:
+            return src
+        if not (src.flags['C_CONTIGUOUS'] or src.flags['F_CONTIGUOUS']):
+            raise NotImplementedError('Only contiguous arrays are currently supported for gpu-gpu transfers.')
+        dst = cupy.empty_like(src)
+        dst.data.copy_from_device_async(src.data, src.nbytes)
```
nicelhc13, line 61: Doesn't it need to specify the stream to be used?
The default stream of this function is CUDA context, and IIRC, this usually means per-device context.
The default value of `cudaMemcpyAsync()` for the stream is also '0.'

insertinterestingnamehere: No. It'll use the current stream. See https://github.com/cupy/cupy/blob/5880b9894b9912a1172e33d7fb3d0d7834eb7832/cupy/cuda/memory.pyx#L433.

nicelhc13 (approved):
LGTM! Thank you!

insertinterestingnamehere:
Great, thanks for reviewing!

PR #48, opened 2021-04-08
Fix Persistent Memory Reservation

### LINKED ISSUE #47: Persistent memory location
I am not sure if I am using persistent memory parameter properly, but my code doesn't work as I expected.

This is parts of the multiGPU blocked cholesky code and usage of the persistent memory parameter: [here](https://gist.github.com/nicelhc13/8b800fd135ae9cdce1ff50b8337310aa#file-gistfile1-txt-L60)

I would like to pre-define/allocate persistent memory for each device before starting the loop.

[This](https://gist.github.com/nicelhc13/8b800fd135ae9cdce1ff50b8337310aa#file-gistfile1-txt-L158) call allocates and append memory for each GPU to the `gpu_arrs`. Therefore, the first dimension of the `gpu_arrs` is memory for each gpu (like gpu_arrs[0] -> gpu0, gpu_arrs[1] -> gpu1, ..)

Then, I pass `gpu_arrs` to the `reserve_persistent_memory()`.

My expectation is each gpu is aware of data owned by them.
(e.g. gpu0 knows gpu_arrs[0].nbytes are residing on its memory)

But, Parla allocates all memory to the current device. 
For example, my usecase allocates all memory on the gpu0.
This is because `reserve_persistent_memory()` makes reference or copies the source data first
and then uses device information of its newly created/referenced array.
[Link](https://github.com/ut-parla/Parla.py/blob/ce1a4d3e284a7bee83dd44a98a9f205d56f317ed/parla/tasks.py#L512)

Therefore, it degraded performance.

For tests, my temporary solution is to replace `view.device` with `amount.device`.

I was wondering if it is right usecase.

insertinterestingnamehere:
Fixes #47. The persistent memory reservation was unintentionally doing all reservations on a single device because of the cupy.asarray call. This removes that.

### DIFF
--- a/parla/tasks.py
+++ b/parla/tasks.py
@@ -501,15 +501,11 @@ def reserve_persistent_memory(amount, device = None):
         memsize = amount
     elif hasattr(amount, '__cuda_array_interface__'):
         import cupy
-        # cupy has the info we need and views are cheap,
-        # so get a view then read off the size/device info.
-        view = cupy.array(amount, copy = False)
-        # TODO: there's actually a storage_size helper routine in
-        # parla.array. We should use it here and move the logic
-        # about buffer protocols there.
-        memsize = view.nbytes
+        if not isinstance(amount, cupy.ndarray):
+            raise NotImplementedError("Currently only CuPy arrays are supported for making space reservations on the GPU.")
+        memsize = amount.nbytes
         if device is None:
-            device = view.device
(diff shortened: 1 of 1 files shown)


insertinterestingnamehere:
This is a pretty minor change, so I'll merge soon unless anyone wants to review in further detail.

Issue #37, opened 2021-04-06
Should Parla tasks/contexts also setup stream handles for numba?


dialecticDolt:
Inspired by the recent discussion on Slack around https://gist.github.com/leofang/4a043e5d94b4702d04fde2b9e7dcebbd 
and passing the current stream to numba kernels. 

Looking forward, would it make sense to provide a local variable within a task for this as well to prevent the user from creating and destroying these stream wrappers themselves?

arthurp:
If there is a way to set the default numba CUDA stream, then Parla should have a component that can set that automatically before entering each task. Then the same stream can be automatically used everywhere you would expect it to be.

dialecticDolt:
From what I'm seeing on this, they follow the CUDA API quite closely which afaik doesn't have a way of setting a default. (aside from the usual two defaults: per-thread and not)

They just have the pattern of passing it into each kernel call.

insertinterestingnamehere:
See also: https://github.com/numba/numba/issues/5137

insertinterestingnamehere:
Actually that's only semi-related. That's about using the per-thread default streams introduced in CUDA 7 in their CUDA module. It's not suggesting interop between numba kernel calls and cupy stream context management idioms.

insertinterestingnamehere:
https://github.com/numba/numba/issues/4797 is semi-related. I'll try creating an issue upstream to see what they say though.

insertinterestingnamehere:
Submitted upstream as https://github.com/numba/numba/issues/4797. We'll see if they can give us some guidance.

insertinterestingnamehere:
Submitted upstream as https://github.com/numba/numba/issues/6921. We'll see if they can give us some guidance.

insertinterestingnamehere:
They suggested using the per-thread default stream as a potential workaround. I initially didn't think that'd work, but since we already are synchronizing the stream at the end of each task and since CUDA's per-thread default stream mode runs those default streams in asynchronous mode, I think that may actually do what we need. The catch is that it requires that libraries be compiled with per-thread default stream enabled. CuPy claims to support this if `CUPY_CUDA_PER_THREAD_DEFAULT_STREAM` is set to 1 (I'm not sure how they do that with the same binary...). HIP doesn't have an equivalent though. Does anyone else have any thoughts on this?

arthurp:
I worry about requiring specific features in the libraries we use. But without VECs it's kind of hard to avoid, so I think this would be reasonable.

[...]

PR #54, opened 2021-04-16
Switch to using the per-thread default stream

insertinterestingnamehere:
Given that https://github.com/numba/numba/pull/6936 is landing soon, this switches us over to using the per-thread default stream in parla. I'll see if I can also get cupy to also support switching to using per-thread-default-stream mode after import since that'll let us get rid of the import order dependence.

The import order issue may cause some existing apps to error out. Further investigation of the implications is pending.


insertinterestingnamehere:
Okay, this works with development versions of cupy and numba (with https://github.com/numba/numba/pull/6936). Once those patches are available in the binaries we can switch to using the per-thread default stream to resolve https://github.com/ut-parla/Parla.py/issues/37.

insertinterestingnamehere:
Note: this does not work with current cupy binaries for some reason. It does work with the development version if I compile it myself. See https://github.com/cupy/cupy/issues/5092 for what details I have. We'll see if the next round of release binaries are okay.

Issue #11, opened 2020-12-17
VECs should be optional


insertinterestingnamehere:
Currently VECs interfere with our hardware topology layer because they have to be imported before numpy, but some other parts of the code may need to use numpy. Really all of the following should be true:
- Parla should be usable without the modified glibc and forwarding libraries if a user doesn't use VECs.
- Trying to use VECs without the supporting libraries should raise an informative error.
- VECs should be usable if the supporting libraries are loaded correctly.

insertinterestingnamehere:
IIRC, we need some cleanup beyond https://github.com/ut-parla/Parla.py/pull/58 to make sure things are properly import-order independent within parla modules.

PR #58, opened 2021-04-20
Fix parla cpu/multiload/numpy dependency cycle.

insertinterestingnamehere:


insertinterestingnamehere:
Fixes https://github.com/ut-parla/Parla.py/issues/11.

insertinterestingnamehere:
This patch appears to work, but I'm certainly still putting it through its paces.

parla/cpu.py
```diff
@@ -1,66 +1,17 @@
 import logging
 from typing import Dict, Collection
 
-# FIXME: This load of numpy causes problems if numpy is multiloaded. So this breaks using VECs with parla tasks.
-#  Loading numpy locally works for some things, but not for the array._register_array_type call.
```
arthurp, line 5: I think this is still an issue with this code. Why did you remove the FIXME.

arthurp (approved):
Fixes the dependency cycle. But not the VEC issue as importing `parla.cpu` outside multiload will still mess things up by importing numpy early.

Otherwise, LGTM.

insertinterestingnamehere:
Right. I think the idiom we want to use for now goes like this:

```
from parla.multiload import multiload

with multiload():
    import numpy

from parla.cpu import cpu
```

So, things work, but multiload has to be imported first. I don't think this is ideal yet, but it'll get things going for the VEC demos again.

parla/cpu.py
insertinterestingnamehere: I put it back in. This is only partially fixed now.

PR #86, opened 2021-09-17
Add tutorial section for data partitioning/movement

bozhiyou:
From the users' perspective, a tutorial on data *partitioning* makes more sense since the data *movement* happens silently and is based on the partitioning.


dialecticDolt:
LGTM, the only things I can think to add would be:
(1) starting with a mapper that partitions over multiple devices 
(2) the overlap/ghost points features of the current partitioners.

sestephens73:
I've merged the request. @bozhiyou if you need to make changes, feel free to make a new merge request, or to commit directly to the tutorial branch (that's what we've been doing and we haven't hit conflicts so far).

PR #95, opened 2022-01-20
Reconstruct scheduler

nicelhc13:
This PR reconstructs the existing scheduler.
Tasks pass through four steps before they are run; 1) spawn, 2) device mapped, 3) ready queue, and 4) device queue.
1) Application spawns tasks.
2) Mapper on the scheduler assigns a device to tasks on the spawned queue.
In this case, task dependencies are not necessarily resolved. This PR creates a separate data movement task for the computation task.
3) If target devices are mapped, the task waits until its dependencies are all done. When the dependencies are done, they would decrease a count on the dependee that specifies the number of remaning higher priority tasks. When the count becomes 0, the task is enqueued onto a ready queue.
4) This PR polls the ready queue and immediately moves tasks on it to the device queue. FYI, this queue will handle and sort task priorities.
5) Lastly, heads of each device queue will run their tasks. This PR assigns a dedicated thread to each device queue and that will run tasks on it. In the future, we will use a launcher (presumably using a dedicated thread) and it will assign a worker thread to a task and will run it.

### DIFF
--- a/examples/quicksort.py
+++ b/examples/quicksort.py
@@ -73,4 +73,5 @@ if __name__ == '__main__':
         quicksort(a)
     end = time.perf_counter()
     print(end - start, "seconds")
+    print("Result:", (a == a_sorted).all())
     assert (a == a_sorted).all()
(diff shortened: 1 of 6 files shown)


sestephens73 (approved):
Looks good to me based on discussion in Slack

PR #96, opened 2022-01-29
feat: single device PArray with automatic datamovement

yinengy:
fix import error

implement __getattr__ and fix type checking

feat: asarray_batch

feat: kmeans_plusplus

Ignore .vscode

fix: typo

feat: add dataflow to spawn and task

feat: support auto_move in task_runtime

fix: missing placement

feat: support index for multi GPU

feat: more APIs about data movement

improvement: using PArray now

refactor: _on_same_device

### DIFF
--- a/parla/dataflow.py
+++ b/parla/dataflow.py
@@ -1,0 +1,26 @@
+"""
+This module implement dataflow tracking.
+Which is used to provide data-aware scheduling,
+and also eager-fashion automatic data movement.
+"""
+
+from typing import Collection, Any
+from itertools import chain
+
+class Dataflow:
+    """
+    The data reference of input/output/inout of a task
+    """
+
+    def __init__(self, input: Collection[Any], output: Collection[Any], inout: Collection[Any]):
+        self._input = input
+        self._output = output
+        self._inout = inout
+
+    def auto_move(self):
+        """
+        Move all data to the current device (of the corresponding tasks).
+        Only PArray is supported.
+        """
+        for array in chain(self._input, self._output, self._inout):
+            array._auto_move()
--- a/parla/dataflow.py
+++ b/parla/dataflow.py
@@ -1,6 +1,6 @@
 """
 This module implement dataflow tracking.
-Which is used to provide data-aware scheduling,
+It is used to provide data-aware scheduling,
 and also eager-fashion automatic data movement.
 """
 
--- a/parla/device.py
+++ b/parla/device.py
@@ -109,8 +109,8 @@ class Device(metaclass=ABCMeta):
         """
         Construct a new Device with a specific architecture.
         """
-        self.architecture = architecture
-        self.index = index
+        self.architecture = architecture  # typically parla.cpu_impl.cpu or parla.cuda.gpu
+        self.index = index  # typically as index of gpu
         self.args = args
         self.kwds = kwds
 
--- a/parla/device.py
+++ b/parla/device.py
@@ -109,8 +109,8 @@ class Device(metaclass=ABCMeta):
         """
         Construct a new Device with a specific architecture.
         """
-        self.architecture = architecture  # typically parla.cpu_impl.cpu or parla.cuda.gpu
-        self.index = index  # typically as index of gpu
+        self.architecture = architecture  # parla.cpu_impl.cpu or parla.cuda.gpu TODO(yineng): more architectures could be added in the future
+        self.index = index  # index of gpu
         self.args = args
         self.kwds = kwds
 
--- a/parla/parray/core.py
+++ b/parla/parray/core.py
@@ -1,3 +1,5 @@
+from __future__ import annotations # To support self references in type checking, must be first line
+
 import numpy
 import cupy
 
@@ -10,7 +12,7 @@ class PArray:
     """Multi-dimensional array on a CPU or CUDA device.
 
     This class is a wrapper around :class:`numpy.ndarray` and :class:`cupy.ndarray`,
-    It is used to support Parla sheduler optimization and automatic data movement
+    It is used to support Parla sheduler optimization and automatic data movement.
 
     Args:
         array: :class:`cupy.ndarray` or :class:`numpy.array` object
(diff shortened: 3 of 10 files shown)


parla/dataflow.py
```diff
@@ -0,0 +1,26 @@
+"""
+This module implement dataflow tracking.
+Which is used to provide data-aware scheduling,
```
nicelhc13, line 3: Which is -> It is

parla/device.py
```diff
@@ -109,8 +109,8 @@ def __init__(self, architecture: "Architecture", index, *args, **kwds):
         """
         Construct a new Device with a specific architecture.
         """
-        self.architecture = architecture
-        self.index = index
+        self.architecture = architecture  # typically parla.cpu_impl.cpu or parla.cuda.gpu
```
nicelhc13, line 112: Please remove typically.
Instead add TODO().
e.g. TODO(yineng): more architectures could be added in the future

parla/device.py
```diff
@@ -109,8 +109,8 @@ def __init__(self, architecture: "Architecture", index, *args, **kwds):
...
+        self.index = index  # typically as index of gpu
```
nicelhc13, line 113: please remove typically.

parla/parray/core.py
```diff
@@ -0,0 +1,458 @@
...
+from parla.cpu_impl import cpu
+from parla.cuda import gpu
+from parla.tasks import get_current_devices
+from parla.device import Device
+
+class PArray:
+    """Multi-dimensional array on a CPU or CUDA device.
+
+    This class is a wrapper around :class:`numpy.ndarray` and :class:`cupy.ndarray`,
+    It is used to support Parla sheduler optimization and automatic data movement
```
nicelhc13, line 13: Add '.'

parla/parray/core.py
```diff
@@ -0,0 +1,458 @@
...
+
+    @property
+    def _on_gpu(self) -> bool:
+        """
+        True if the array is on GPU
+        """
+        return isinstance(self.array, cupy.ndarray)
+
+    @property
+    def _index(self) -> int:
```
nicelhc13, line 29: rename it to `_dev_index()`?

parla/parray/core.py
```diff
@@ -0,0 +1,458 @@
...
+        """
+        Index of Current Device
```
nicelhc13, line 31: Please update this to:

Return -1 if the current device is CPU.
Otherwise return GPU ID.

parla/parray/core.py
```diff
@@ -0,0 +1,458 @@
...
+        elif device.architecture == cpu:
+            self._to_cpu()
+
+    def _to_current_gpu(self) -> None:
+        """
+        Move the array to current GPU, do nothing if already on the device.
+        """
+        self.array = cupy.asarray(self.array)  # asarray by default copy to current device
+
+    def _to_gpu(self, index: int) -> None:
```
nicelhc13, line 63: Do we need this function in the future?

nicelhc13: I meant you pointed me out that moving to the current gpu does not need id specification and i agreed with you :-)

parla/parray/core.py
```diff
@@ -0,0 +1,458 @@
...
+        Move the array to GPU, do nothing if already on the device.
+        `index` is the index of GPU copied to
+        """
+        if self._index == index:
+            return
+
+        with cupy.cuda.Device(index):
+            self.array = cupy.asarray(self.array)
+
+    def _to_cpu(self) -> None:
```
nicelhc13, line 74: same with the above

parla/parray/core.py
```diff
@@ -0,0 +1,458 @@
...
+        """
+        Move the array to CPU, do nothing if already on CPU.
+        """
+        if not self._on_gpu:
+            return
+        self.array = cupy.asnumpy(self.array)
+
+    def _on_same_device(self, other) -> bool:
```
nicelhc13, line 82: other -> `other: PArray`

parla/parray/from_data.py
```diff
@@ -0,0 +1,148 @@
+import numpy
+import cupy
+
+from .core import PArray
+
+
+def array(object, dtype=None, copy=True, order='K', subok=False, ndmin=0, like=None, on_gpu=False):
+    """
+    Creates an Parla array specific device (CPU by default).
```
nicelhc13, line 9: Create a Parla array on the specific device?

parla/tasks.py
```diff
@@ -341,7 +342,10 @@ def spawn(taskid: Optional[TaskID] = None, dependencies = (), *,
           placement: Union[Collection[PlacementSource], Any, None] = None,
           ndevices: int = 1,
           tags: Collection[Any] = (),
-          data: Collection[Any] = None
+          data: Collection[Any] = None,
+          input: Collection[Any] = (),
+          output: Collection[Any] = (),
+          inout: Collection[Any] = ()
```
nicelhc13, line 348: just curious as a Python novice. What is difference between () and None? If they are the same, how about just matching to one?

nicelhc13 (approved):
Implementation-wise, it looks good to me.
I mostly wrote something regarding code qualities.
Please check my comments and fix them if you agree with me.

parla/parray/core.py
yinengy: There is a "_to_current_gpu()", which is by default "_to_gpu(current_id)"
I leave this method since it may be used in the future to move data into a specific device rather than current one.

parla/parray/core.py
yinengy: same, reserved for future use.

parla/tasks.py
yinengy: () make it a empty tuple by default, which is iterable. 
There is a chain(self._input, self._output, self._inout) in dataflow.py:25 which cat them and iterate over them. An error will be raised it one of them is none.

yinengy:
Thank you for reviewing, it is resolved and merged now.

PR #97, opened 2022-02-01
fix: hide cupy in system without GPU

yinengy:


parla/parray/core.py
```diff
@@ -1,13 +1,17 @@
...
+import numpy
+try:  # if the system has no GPU
+    import cupy
+    from parla.cuda import gpu
+except (ImportError, AttributeError):
+    cupy = numpy  # work around of cupy.ndarray
+    gpu = None
+
 class PArray:
     """Multi-dimensional array on a CPU or CUDA device.
```
nicelhc13, line 17: Could you please leave a comment that PArray only considers numpy or cupy array? This is because line 34 assumes this and therefore, lefting comments looks safe to me just in case that more array types are used.

nicelhc13 (approved):
Looks good to me. It would be great if only one comment is there; PArray always assume cupy or numpy array. Because otherwise it becomes dangerous.

yinengy:
Resolved, thanks!

PR #99, opened 2022-02-04
Launch phase uses device queues

sestephens73:
Prior to this, there were no queue structures associated with devices. Rather, WorkerThreads had queues (of depth 2). The schedudler would queue up work onto the thread and threads would run tasks that had been assigned to them.

Since threads shouldn't be treated as finite resources, but devices should, we now queue up work on devices instead. (This is the model we discussed in the past few meetings - see attached image).
scheduler

[5ecad4b](https://github.com/ut-parla/Parla.py/pull/99/commits/5ecad4bb16fb0b428355dc87f9d7a3eb1c0f77c5) is just a bunch of comments and a name change (allocation queue -> ready queue). Not much need for review here.

[bbcfffa](https://github.com/ut-parla/Parla.py/pull/99/commits/bbcfffad2834a5d9fb7669ea0dcafe63dfd64637) is just a .gitignore update because I use Vim and didn't want to push *.swp files to the repo.

[97350f8](https://github.com/ut-parla/Parla.py/pull/99/commits/97350f80965e4c3a14f3a59e36d2c438f6a0ff44) is the main set of changes. I recommend starting on like 1271 when reviewing. I've reworked major portions of the WorkerThread class to take only one task rather than acting as a queue. I've updated the scheduler launch function to iterate device queues and pop work as necessary. I've also added a dict storing whether or not each device is busy as a temporary workaround.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -837,7 +837,7 @@ class WorkerThread(ControllableThread, SchedulerContext):
                         self._status = "Running Task {}".format(self.task)
                         self.task.run()
                         self._remove_task()
-                        self.scheduler.free_thread(self)
+                        self.scheduler.append_free_thread(self)
                     # Thread wakes up without a task (should only happen at end of program)
                     elif not self.task and self._should_run:
                         raise WorkerThreadException("%r woke up without a valid task.", self)
@@ -857,7 +857,6 @@ class ResourcePool:
     _multiplier: float
     _monitor: threading.Condition
     _devices: Dict[Device, Dict[str, float]]
-
 
     # Resource pools track device resources. Environments are a separate issue and are not tracked here. Instead,
     # tasks will consume resources based on their devices even though those devices are bundled into an environment.
@@ -965,10 +964,9 @@ class Scheduler(ControllableThread, SchedulerContext):
     _worker_threads: List[WorkerThread]
     _free_worker_threads: Deque[WorkerThread]
     _available_resources: ResourcePool
-    period: float # TODO: Figure out what this is for
+    period: float
 
-    def __init__(self, environments: Collection[TaskEnvironment], n_threads: int = None, period: float = 0.01,
-                 max_worker_queue_depth: int = 2):
+    def __init__(self, environments: Collection[TaskEnvironment], n_threads: int = None, period: float = 0.01):
         # ControllableThread: __init__ sets it to run
         # SchedulerContext: No __init__
         super().__init__()
@@ -986,13 +984,9 @@ class Scheduler(ControllableThread, SchedulerContext):
         # Start with one count that is removed when the scheduler is "exited"
         self._active_task_count = 1
 
-        # TODO: Delete this
-        self.max_worker_queue_depth = max_worker_queue_depth
-
-        # TODO: Figure out what this is for
+        # Period scheduler sleeps between loops (see run function)
         self.period = period
 
-        # TODO: Figure out what this is for
         self._monitor = threading.Condition(threading.Lock())
 
         # Track, allocate, and deallocate resources (devices)
@@ -1011,10 +1005,6 @@ class Scheduler(ControllableThread, SchedulerContext):
         # tasks spawn as it runs
         self._spawned_task_queue = deque()
         self._new_spawned_task_queue = deque()
-
-        # TODO: Delete these since they're not used I think
-        self._mapped_task_queue = deque()
-        self._new_mapped_task_queue = deque()
 
         # This is where tasks go when they have been mapped and their
         # dependencies are complete, but they have not been scheduled.
@@ -1058,7 +1048,7 @@ class Scheduler(ControllableThread, SchedulerContext):
             # TODO: Should combine all of them into a single exception.
             raise self._exceptions[0]
 
-    def free_thread(self, thread: WorkerThread):
+    def append_free_thread(self, thread: WorkerThread):
         with self._monitor:
             self._free_worker_threads.append(thread)
 
@@ -1302,7 +1292,7 @@ class Scheduler(ControllableThread, SchedulerContext):
                 self._schedule_tasks()
                 self._launch_tasks()
                 logger.debug("[Scheduler] Sleeping!")
-                time.sleep(1) # TODO: Delete this. Why the fuck doesn't this release the GIL.
+                time.sleep(self.period)
                 logger.debug("[Scheduler] Awake!")
 
         except Exception:
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -973,45 +973,75 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
     return lst
 
 class Scheduler(ControllableThread, SchedulerContext):
+    # See __init__ function below for comments on the functionality of these members
     _environments: TaskEnvironmentRegistry
-    # Worker threads for each device.
     _worker_threads: List[WorkerThread]
     _available_resources: ResourcePool
-    period: float
-    max_worker_queue_depth: int
+    period: float # TODO: Figure out what this is for
+    max_worker_queue_depth: int # TODO: Delete this
 
     def __init__(self, environments: Collection[TaskEnvironment], n_threads: int = None, period: float = 0.01,
                  max_worker_queue_depth: int = 2):
+        # ControllableThread: __init__ sets it to run
+        # SchedulerContext: No __init__
         super().__init__()
+
         # TODO(lhc): for now, assume that n_threads is always None.
         #            Each device needs a dedicated thread.
         n_threads = sum(d.resources.get("vcus", 1) for e in environments for d in e.placement)
+
+        # TODO: Figure out what these are for
         self._environments = TaskEnvironmentRegistry(*environments)
+
+        # Empty list for storing reported exceptions at runtime
         self._exceptions = []
-        self._active_task_count = 1 # Start with one count that is removed when the scheduler is "exited"
+
+        # Start with one count that is removed when the scheduler is "exited"
+        self._active_task_count = 1
+
+        # TODO: Delete this
         self.max_worker_queue_depth = max_worker_queue_depth
+
+        # TODO: Figure out what this is for
         self.period = period
+
+        # TODO: Figure out what this is for
         self._monitor = threading.Condition(threading.Lock())
-        self._allocation_queue = deque()
-        # Spawned queue should consist of two levels, current and new.
-        # New spawned tasks or tasks failed to schedule are always
-        # enqueued on the new queue.
-        # Tasks which the scheduler will try to schedule at the current
+
+        # Spawned task queues
+        # Tasks that have been spawned but not mapped are stored here.
+        # Tasks are removed once they are mapped.
+        # Spawned queue consists of two levels, current and new.
+        # Newly spawned tasks or tasks which fail to schedule are always
+        # enqueued on the "new" queue.
+        # When the mapper runs, it moves all tasks from the "new" to the "current" queue.
+        # Tasks which the mapper will try to map at the current
         # iteration are always dequeued from the current queue.
-        # For each iteration, elements on the new queue are moved and
-        # appended to the current queue.
-        # Through the two level queues, the scheduler could avoid tailing elements,
-        # and move to the next super step for the mapped tasks.
+        # This implementation is simple and avoids a long-running mapper in the case where new
+        # tasks spawn as it runs
         self._spawned_task_queue = deque()
         self._new_spawned_task_queue = deque()
+
+        # TODO: Delete these since they're not used I think
         self._mapped_task_queue = deque()
         self._new_mapped_task_queue = deque()
-        self._device_queues = []
+
+        # This is where tasks go when they have been mapped and their
+        # dependencies are complete, but they have not been scheduled.
+        self._ready_queue = deque()
+
+        # The device queues where scheduled tasks go to be launched from
         self._device_queues = [deque() for n in range(0, n_threads)]
+
+        # Track, allocate, and deallocate resources (devices)
         self._available_resources = ResourcePool(multiplier=1.0)
+
+        # Worker threads for each device (probably removing these)
         self._worker_threads = [WorkerThread(self, i) for i in range(n_threads)]
         for t in self._worker_threads:
             t.start()
+
+        # Start the scheduler thread (likely to change later)
         self.start()
 
     @property
@@ -1091,7 +1121,7 @@ class Scheduler(ControllableThread, SchedulerContext):
         """Enqueue a task on the resource allocation queue.
            Note that this enqueue has no data race.
         """
-        self._allocation_queue.appendleft(task)
+        self._ready_queue.appendleft(task)
 
     def _dequeue_task(self, timeout=None) -> Optional[Task]:
         """Dequeue a task from the resource allocation queue.
@@ -1099,7 +1129,7 @@ class Scheduler(ControllableThread, SchedulerContext):
         while True:
             try:
                 if self._should_run:
-                    return self._allocation_queue.pop()
+                    return self._ready_queue.pop()
                 else:
                     return None
             except IndexError:
@@ -1258,7 +1288,7 @@ class Scheduler(ControllableThread, SchedulerContext):
 
     def dump_status(self, lg=logger):
         lg.info("%r:\n%r\navailable: %r", self,
-                self._allocation_queue, self._available_resources)
+                self._ready_queue, self._available_resources)
         w: WorkerThread
         for w in self._worker_threads:
             w.dump_status(lg)
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -1,4 +1,3 @@
-import abc
 import logging
 import random
 from abc import abstractmethod, ABCMeta
@@ -7,12 +6,9 @@ from contextlib import contextmanager
 import threading
 import time
 from itertools import combinations
-from numbers import Number
-from threading import Thread, Condition
 from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSet, Iterable, TypeVar, Deque
-import concurrent.futures
 
-from parla.device import get_all_devices, Device, Architecture
+from parla.device import get_all_devices, Device
 from parla.environments import TaskEnvironmentRegistry, TaskEnvironment
 from parla.dataflow import Dataflow
 
@@ -33,7 +29,7 @@ _ASSIGNMENT_FAILURE_WARNING_LIMIT = 32
 TaskAwaitTasks = namedtuple("AwaitTasks", ("dependencies", "value_task"))
 
 
-class TaskState(object, metaclass=abc.ABCMeta):
+class TaskState(object, metaclass=ABCMeta):
     __slots__ = []
 
     @property
@@ -95,7 +91,7 @@ class TaskException(TaskState):
 ResourceDict = Dict[str, Union[float, int]]
 
 
-class ResourceRequirements(object, metaclass=abc.ABCMeta):
+class ResourceRequirements(object, metaclass=ABCMeta):
     __slots__ = ["resources", "ndevices", "tags"]
 
     tags: FrozenSet[Any]
@@ -749,7 +745,7 @@ def get_devices() -> Collection[Device]:
     return _scheduler_locals.environment.placement
 
 
-class ControllableThread(Thread, metaclass=ABCMeta):
+class ControllableThread(threading.Thread, metaclass=ABCMeta):
     _should_run: bool
     _monitor: threading.Condition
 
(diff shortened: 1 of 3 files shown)


parla/task_runtime.py
```diff
@@ -9,7 +9,8 @@
 from itertools import combinations
 from numbers import Number
 from threading import Thread, Condition
-from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSet, Iterable, TypeVar
+from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSet, Iterable, TypeVar, Deque
```
sestephens73, line 12: added Deque here so I can do some type annotations later

parla/task_runtime.py
```diff
@@ -9,7 +9,8 @@
...
+import concurrent.futures
```
sestephens73, line 13: Whoops, I need to delete this line! I plan to do a "cleanup" commit later after we merge some stuff, so I'll take care of it then.

parla/task_runtime.py
```diff
@@ -379,6 +380,8 @@ def run(self):
                 # the scheduler.
                 for d in self.req.devices:
                     ctx.scheduler._available_resources.deallocate_resources(d, self.req.resources)
+                    # TODO: This is part of the easy launcher hack, fix it
+                    ctx.scheduler._available_resources._occupancy_dict[d] = False
```
sestephens73, line 384: I currently keep a (hacky) dict in the scheduler listing which resources are free or not. Later I will update this to properly use the ResourcePool.

parla/task_runtime.py
```diff
@@ -760,17 +766,16 @@ def stop(self):
...
     def __init__(self, scheduler, index):
         super().__init__()
         self._monitor = threading.Condition(threading.Lock())
         self.index = index
         self._scheduler = scheduler
-        # Use a deque to store local tasks (a high-performance implementation would a work stealing optimized deque).
-        # In this implementation the right is the "local-end", so append/pop are used by this worker and
-        # appendleft/popleft are used by the scheduler or other workers.
-        self._queue = deque()
+        self.task = None
```
sestephens73, line 778: replace the queue with a single task

parla/task_runtime.py
```diff
@@ -783,51 +788,6 @@ def incr_active_tasks(self):
...
-            except IndexError:
-                return None
-
-    def _push_task(self, task):
-        """Push a local task on the queue head.
-        """
-        with self._monitor:
-            self._queue.append(task)
-            self._monitor.notify()
-
```
sestephens73, line 830: deleted a crapton of code we no longer need since WorkerThreads don't need queues

parla/task_runtime.py
```diff
@@ -847,24 +807,39 @@ def enqueue_task(self, task):
...
+                        with self._monitor:
+                            self._monitor.wait()
+                        print("[WorkerThread ", self.index,"] Waking up.", sep='')
+                    if not self.task:
+                        break # TODO: Properly raise exception here
+                    print(f"[WorkerThread {self.index}] Starting:", self.task.name, sep='')
+                    self._status = "Running Task {}".format(self.task)
+                    self.task.run()
+                    self._remove_task()
+                    self.scheduler.free_thread(self)
```
sestephens73, line 842: This code is basically copied from here https://github.com/ut-parla/Parla.py/pull/99/commits/97350f80965e4c3a14f3a59e36d2c438f6a0ff44#diff-69c8f130075278971cac4616ec977714100e803d919c4a099059a572fb28201dL796-L812

but it supports a single task instead of a queue

parla/task_runtime.py
```diff
@@ -976,9 +961,9 @@ class Scheduler(ControllableThread, SchedulerContext):
     # See __init__ function below for comments on the functionality of these members
     _environments: TaskEnvironmentRegistry
     _worker_threads: List[WorkerThread]
+    _free_worker_threads: Deque[WorkerThread]
```
sestephens73, line 964: `_worker_threads` refers to all of them. `_free_worker_threads` is just the free ones at any given time. Busy threads are popped from this deque, and when they finish their task they reappend themselves to the free deque.

parla/task_runtime.py
```diff
@@ -1008,6 +993,9 @@ def __init__(self, environments: Collection[TaskEnvironment], n_threads: int = N
         # TODO: Figure out what this is for
         self._monitor = threading.Condition(threading.Lock())
 
+        # Track, allocate, and deallocate resources (devices)
+        self._available_resources = ResourcePool(multiplier=1.0)
+
```
sestephens73, line 998: this is just moved from line 1037 on the left

parla/task_runtime.py
```diff
@@ -1031,15 +1019,13 @@ def __init__(self, environments: Collection[TaskEnvironment], n_threads: int = N
         self._ready_queue = deque()
 
         # The device queues where scheduled tasks go to be launched from
-        self._device_queues = [deque() for n in range(0, n_threads)]
-
-        # Track, allocate, and deallocate resources (devices)
-        self._available_resources = ResourcePool(multiplier=1.0)
+        self._device_queues = {dev: deque() for dev in self._available_resources.get_resources()}
```
sestephens73, line 1022: these are the new device queues!

parla/task_runtime.py
```diff
@@ -1255,23 +1254,54 @@ def run_mapper(self):
...
         try: # Catch all exception to report them usefully
+            i = 0
             while self._should_run:
-                self.run_mapper()
-                self.run_deviceq_mapper()
+                self._map_tasks()
+                self._schedule_tasks()
+                self._launch_tasks()
+                print("[Scheduler] Sleeping!")
+                time.sleep(1) # TODO: Delete this. Why the fuck doesn't this release the GIL.
```
sestephens73, line 1302: we need to adjust this to sleep for a dynamically set `period`

sestephens73: also, ignore my angry comment. It does release the GIL. I'm just dumb.

parla/task_runtime.py
```diff
@@ -1255,23 +1254,54 @@ def run_mapper(self):
...
+                # Only works for single-device tasks, and only one task at a time on each device
+                # TODO: Make this not terrible
+                if self._available_resources._occupancy_dict[dev] == False: # if not occupied
+                    if len(queue) > 0: # if there are tasks on the queue
+                        task = queue.pop() # grab a task
+                        worker = self._free_worker_threads.pop() # grab a worker
+                        print(f"[Scheduler] Launching {task.name} on WorkerThread {worker.index}")
+                        self._available_resources._occupancy_dict[dev] = True # mark the device as occupied
+                        worker.assign_task(task) # assign the task to the worker (this notifies the worker's monitor)
+                        print(f"[Scheduler] Launched {worker.task.name}")
```
sestephens73, line 1291: `_launch_tasks` is the main new thing in this commit and all the other changes are there to just make this work.

parla/task_runtime.py
```diff
@@ -1225,23 +1254,54 @@ def run_mapper(self):
...
     def run(self) -> None:
         # noinspection PyBroadException
         try: # Catch all exception to report them usefully
+            i = 0
             while self._should_run:
-                self.run_mapper()
-                self.run_deviceq_mapper()
+                self._map_tasks()
+                self._schedule_tasks()
+                self._launch_tasks()
```
dialecticDolt, line 1300: One thing I'm concerned about reading this is overhead and/or stalling. 

iiuc (if I understand correctly) the execution is: 
- A group of tasks are mapped to devices
- A group of tasks are scheduled (assigned priority in device queues corresponding to their mapping) and are ready to run
- All devices are iterated over, ready tasks are grabbed, and attempted to be run on worker threads. A worker thread can only have 1 task assigned to it (the currently running task). 

Assuming the above understanding, and that I'm not missing any work stealing, if any workers finish their work they need to wait until the next time L1300 is called to have any work to do. This can be after potentially expensive and time consuming "Map" and "Schedule" phases (depending on policy).

parla/task_runtime.py
```diff
@@ -1225,23 +1254,54 @@ def run_mapper(self):
...
+                print("[Scheduler] Sleeping!")
+                time.sleep(1) # TODO: Delete this. Why the fuck doesn't this release the GIL.
```
dialecticDolt, line 1302: Should this wake on events?

parla/task_runtime.py
```diff
@@ -847,24 +807,39 @@ def enqueue_task(self, task):
...
-                    self._status = "Running Task {}".format(task)
-                    task.run()
+                    if not self.task:
+                        logger.debug("Blocking for a task: %r (%s)", self, self._monitor)
+                        print("[WorkerThread ", self.index,"] Blocking for a task.", sep='')
+                        with self._monitor:
+                            self._monitor.wait()
+                        print("[WorkerThread ", self.index,"] Waking up.", sep='')
+                    if not self.task:
+                        break # TODO: Properly raise exception here
```
dialecticDolt, line 837: I'm not sure how 836 can be hit, assuming the thread is waiting for work?

parla/task_runtime.py
dialecticDolt: (example: a worker finishing, etc)

parla/task_runtime.py
dialecticDolt: On the other end, if this loop is very running quickly, then the "mapping" and "scheduling" are likely never populated and batches aren't considered at a time. (although this can be fixed by having a threshold maybe? Not sure what the best pattern is here)

parla/task_runtime.py
```diff
@@ -305,7 +306,7 @@ def _complete_dependency(self):
         with self._mutex:
             self._remaining_dependencies -= 1
             self._check_remaining_dependencies()
-            print("Computation task, ", self.name, "'s one dependency is completed. [remaining:", \
+            print(f"[Task {self.name}] Data movement dependency completed. [remaining:", \
```
nicelhc13, line 309: You can remove this

nicelhc13: Or we can replace it with logger to toggle.

parla/task_runtime.py
```diff
@@ -483,8 +486,8 @@ def _complete_dependency(self):
         with self._mutex:
             self._remaining_dependencies -= 1
             self._check_remaining_dependencies()
-            print("Data-movement task, ", self.name, "'s one dependency is " \
-                  "completed. [remaining:", self._remaining_dependencies, sep='')
+            #print("Data-movement task, ", self.name, "'s one dependency is " \
```
nicelhc13, line 489: You can remove  this

parla/task_runtime.py
nicelhc13: I don't think so since this thread will run the task..

nicelhc13: Was there any reason to put this if-condition?

parla/task_runtime.py
```diff
@@ -881,6 +856,7 @@ class ResourcePool:
     _monitor: Condition
     _devices: Dict[Device, Dict[str, float]]
 
+
```
nicelhc13, line 859: remove?

parla/task_runtime.py
```diff
@@ -973,45 +958,76 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
...
 class Scheduler(ControllableThread, SchedulerContext):
+    # See __init__ function below for comments on the functionality of these members
     _environments: TaskEnvironmentRegistry
-    # Worker threads for each device.
     _worker_threads: List[WorkerThread]
+    _free_worker_threads: Deque[WorkerThread]
     _available_resources: ResourcePool
-    period: float
-    max_worker_queue_depth: int
+    period: float # TODO: Figure out what this is for
```
nicelhc13, line 966: now we don't need this. This was for re-enqueueing delay. If unassigned allocation was not enough, scheduler re-enqueue the task to the queue after this period time.

parla/task_runtime.py
```diff
@@ -973,45 +958,76 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
...
         self._environments = TaskEnvironmentRegistry(*environments)
+
+        # Empty list for storing reported exceptions at runtime
         self._exceptions = []
-        self._active_task_count = 1 # Start with one count that is removed when the scheduler is "exited"
+
+        # Start with one count that is removed when the scheduler is "exited"
+        self._active_task_count = 1
+
+        # TODO: Delete this
```
nicelhc13, line 987: why not just delete this?

parla/task_runtime.py
```diff
@@ -973,45 +958,76 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
...
         self.max_worker_queue_depth = max_worker_queue_depth
+
+        # TODO: Figure out what this is for
```
nicelhc13, line 990: we can remove it

parla/task_runtime.py
```diff
@@ -973,45 +958,76 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
...
         self.period = period
+
+        # TODO: Figure out what this is for
```
nicelhc13, line 993: Why is it questionable? I think this is necessary. For example, any thread can spawn task and it causes race condition.

parla/task_runtime.py
```diff
@@ -973,45 +958,76 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
...
+        # enqueued on the "new" queue.
+        # When the mapper runs, it moves all tasks from the "new" to the "current" queue.
+        # Tasks which the mapper will try to map at the current
         # iteration are always dequeued from the current queue.
-        # For each iteration, elements on the new queue are moved and
-        # appended to the current queue.
-        # Through the two level queues, the scheduler could avoid tailing elements,
-        # and move to the next super step for the mapped tasks.
+        # This implementation is simple and avoids a long-running mapper in the case where new
+        # tasks spawn as it runs
```
nicelhc13, line 1009: Cool :-)

parla/task_runtime.py
```diff
@@ -973,45 +958,76 @@ def shuffled(lst: Iterable[_T]) -> List[_T]:
...
         self._spawned_task_queue = deque()
         self._new_spawned_task_queue = deque()
+
+        # TODO: Delete these since they're not used I think
```
nicelhc13, line 1013: Yes, please delete them.

nicelhc13: My local version also removed these.

parla/task_runtime.py
```diff
@@ -1039,6 +1055,10 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             # TODO: Should combine all of them into a single exception.
             raise self._exceptions[0]
 
+    def free_thread(self, thread: WorkerThread):
```
nicelhc13, line 1058: replace this with append_free_thread()

parla/task_runtime.py
```diff
@@ -1071,7 +1091,9 @@ def _dequeue_spawned_task(self) -> Optional[Task]:
             #            when that will be popped?
             #            scheduler is busy waiting this queue?
             try:
-                return self._spawned_task_queue.pop()
+                task = self._spawned_task_queue.pop()
```
nicelhc13, line 1094: why don't you revert this back? (guess you used this for debugging)

parla/task_runtime.py
```diff
@@ -1091,15 +1113,17 @@ def enqueue_task(self, task: Task):
...
+        self._ready_queue.appendleft(task)
 
     def _dequeue_task(self, timeout=None) -> Optional[Task]:
         """Dequeue a task from the resource allocation queue.
         """
         while True:
             try:
                 if self._should_run:
-                    return self._allocation_queue.pop()
+                    task = self._ready_queue.pop()
```
nicelhc13, line 1124: same as above

parla/task_runtime.py
```diff
@@ -1225,23 +1254,54 @@ def run_mapper(self):
...
+                self._device_queues[d].append(task)
+
+    def _launch_tasks(self):
+        """ Iterate through free devices and launch tasks on them
+        """
+        print("[Scheduler] Launch Phase")
+        with self._monitor:
+            for dev, queue in self._device_queues.items():
+                # Make sure there's an available WorkerThread
+                if len(self._free_worker_threads) == 0:
```
nicelhc13, line 1278: can't we move this to outside of for loop?
can avoid unnecessary instructions.

parla/task_runtime.py
nicelhc13: this looks good!

parla/task_runtime.py
nicelhc13: You understood correctly, IIUC and this worker thread pool causes GIL overheads compared to before. But the main point of the thread pool is to support multi-device task. If one thread per one device is, it is unclear how to run multi-devices task. If the GIL is the real problem, we need to revisit our previous dicussion about C++ back-end. 

Mapping and scheduling would not be populated. Threshold of length looks good to me. 

Your points make sense to me. I will list them on the sheet.

@sestephens73  Do you agree? or any suggestion?

parla/task_runtime.py
nicelhc13: If it works, please remove the comment. I guess we tried to avoid sleep, didn't we? (This is real question)
How about using queue length before calling any function as Will suggested?

parla/task_runtime.py
nicelhc13: You mean, scheduler is sleeping until any thread is available, right?
That looks good to me. Then, we can also exploit task batching.

nicelhc13 (approved):
This looks good to me. Please check my comments.
(mostly, please remove unnecessary prints)

I think @dialecticDolt 's points are very good. I listed his comments on our sheet.
I think his suggestion about event and queue threshold on scheduler looks doable. If you also agree, please include that update on this PR too. If it is not, please let me know. I will also list them too.

parla/task_runtime.py
sestephens73: @dialecticDolt You're correct. This is why we've split the scheduler into three distinct phases: mapping, scheduling, and launching. In the (near) future when we're discussing policies, we can easily run the phases in different periods or loops or orders, etc. Launch can be set to run with a fast period while schedule and map have slower ones, for example. I think this would address your concerns?

@nicelhc13 
> this worker thread pool causes GIL overheads compared to before

I don't think so...how come?

parla/task_runtime.py
sestephens73: It *probably* should wake on events. Mattan and Chris suggested that will likely be the case, but instead of over-engineering it, just start with a scheduler that runs with some (possibly dynamic, adjustable) period.

parla/task_runtime.py
sestephens73: I think we may still end up using `period` to set a dynamically adjustable period on the scheduler loop though? I guess that's TBD. No harm removing it for now.

parla/task_runtime.py
nicelhc13: @sestephens73  After I think about this more, I am now thinking that it would not be an overhead compared to before.

In the past (when we have one thread per device Q), scheduler distributes tasks onto device Qs, each device Q has a dedicated thread, and they pop one head task.

Now we have thread pool and threads on the thread pool will try to get task.

I thought that in thread pool case, threads will compete to get task, it would be one more data race or phase, and therefore, it could cause more GIL invocations.

But this task is distributed through one scheduler thread by thread indexing, which is similar with a dedicated thread.

Sorry!

parla/task_runtime.py
sestephens73: See https://github.com/ut-parla/Parla.py/pull/99#discussion_r801852437

I'm currently leaving it as a sleep until we determine we need a different mechanism (which we probably will, but we don't know that yet).

parla/task_runtime.py
sestephens73: It's for error checking in case a thread somehow wakes up without a valid task to run. I've fixed this in a commit I'm about to push. You're correct that it *should* never be hit.

parla/task_runtime.py
sestephens73: No, because on each loop iteration we consume another worker thread. So we have to make sure there's still a free one every loop.

Perhaps if we run out of worker threads we should spawn new ones?

parla/task_runtime.py
```diff
@@ -57,7 +57,10 @@ def clear_dependencies(self):
         self.dependencies = None
 
     def __repr__(self):
-        return "TaskRunning({}, {}, {})".format(self.func.__name__, self.args, self.dependencies)
+        if self.func:
+            return "TaskRunning({}, {}, {})".format(self.func.__name__, self.args, self.dependencies)
+        else:
+            return "Functionless task"
```
nicelhc13, line 63: This is good lol

sestephens73: I was using the logger for debugging and it kept complaining to me about trying to get `None.__name__` haha

PR #100, opened 2022-02-04
feat: support coherence protocol on PArray for multi devices

yinengy:


nicelhc13:
I didn't check your codes yet, but could you please add unit tests?
Just simple operations on parray are fine.
The reason why I am suggesting is this PArray is core layout for Parla runtime and as it grows up with runtime, debugging would become very hard.

yinengy:
> The reason why I am suggesting is this PArray is core layout for Parla runtime and as it grows up with runtime, debugging would become very hard.

Yeah, I cloud add some unit test for MSI protocol later. This PR is already tested with kmeansplusplus.py so I think it may works the same as simple unit test. If there is still hidden bugs, I think it requires more complicated applications to find out.

parla/parray/core.py
```diff
@@ -14,6 +16,241 @@
...
+        if operator_state == self.MODIFIED:  # already have the right to write data
+            return [MemoryOperation.noop()]
+        else:
+            if self._overall_state == self.INVALID:  # the system doesn't hold a copy before
+                self._coherence_states[operator] = self.MODIFIED if do_write else self.SHARED
+                self._owner = operator
+                self._overall_state = self.MODIFIED if do_write else self.SHARED
+                return [MemoryOperation.noop()]
+            else:  # the system already hold a copy
+                # evict others
```
dialecticDolt: Just a small comment: the evict others pattern is used a lot, maybe it can be refactored into a single function (use evict from below)?

parla/parray/coherence.py
```diff
@@ -0,0 +1,241 @@
...
+    SHARED = 1
+    MODIFIED = 2
+
+    def __init__(self, init_owner: int):
+        """
+        Args:
+            init_owner: the owner of the first copy in the system
+        """
+        self._coherence_states = {init_owner: self.SHARED}  # states of each device
+        self._owner = init_owner  # owner id when MODIFIED / smallest valid device id when SHARED
```
dialecticDolt, line 52: How is owner used?

dialecticDolt (approved):
I think auto-move should register and update Coherence information, but otherwise LGTM on a quick read.

dialecticDolt:
@yinengy Almost all of the coherence class (and maybe parts of parray) is going to need to be made threadsafe, what are the current plans for that?

dialecticDolt:
I'm not sure the proposed two solutions to the slicing problem are mutually exclusive.

 I don't know if its what we want but here's a possible feature wishlist of how it behaves:
 
  if a view onto the parray memory is created:
 - it should be able to be copied to new devices separately, without pulling the whole parent parray
 - if copied to an old device that already has the parent parray, only that segment could be updated
- if its updated it would likely need to propagate any changes back to the parent parray

parla/parray/coherence.py
yinengy: owner the where the data should be loaded from. If there are multiple devices share a copy (SHARED state), the smallest one (so likely to be CPU), will becomes the owner. And later other INVALID device will read from it.

yinengy:
> @yinengy Some all of the coherence class (and maybe parts of parray) is going to need to be made threadsafe, what are the current plans for that?
> 
> I think it only matters for read() changing the owner at the moment.
> 
> But for the far future: if the plan ever changes to add versions, and reference out-of-date copies, like Mattan mentioned today: .items() can change non-safely if an IN: task registers and makes a new copy of an old-version during an update or evict execution to create a new-version, etc.

I am also thinking about threadsafe. But I just don't know how many threads will be used by parla scheduler. if there is only 1 dedicated thread for sheduler, it means the protocol info will always updated by only one thread.

But if threadsafe is a issues in the future. One easy solution maybe let each PArray keep a lock. When do write and read, grab the lock first.

yinengy:
> I'm not sure the proposed two solutions to the slicing problem are mutually exclusive.
> 
> I don't know if its what we want but here's a possible feature wishlist of how it might behave: I think we should decide on expected behavior before thinking about underlying implementation.
> 
> if a view onto the parray memory is created:
> 
> * it should be able to be copied to new devices separately, without pulling the whole parent parray
> * if copied to an old device that already has the parent parray, only that segment could be updated
> * if its updated it would likely need to propagate any changes back to the parent parray

I agree with you. If it is a view, only the part of the data will be moved into the device. I need to write some code to check if this it could be done in a clean way.

parla/parray/core.py
yinengy: The reason I make it a separate one is because the detail is slightly different (the if statement). The Coherence.evict() is not used currently. It is reserved for future use (for example, scheduler could free up GPU memory by call evict on some data)

dialecticDolt:
> 

Agree on lock, although I think `auto_move` is called by data-movement tasks, which are scheduled to run on any thread

yinengy:
@bozhiyou I think there is minor changes need to be made on scheduler since PArray now has new API, I think what you need is using `_auto_move` to move data. The `do_write` arguments should be set based on whether the data marked as IN or as INOUT/OUT.

yinengy:
> > 
> 
> Agree on lock, although I think `auto_move` is called by data-movement tasks, which are scheduled to run on any thread

Isn't the movement task run in sequential? @nicelhc13

sestephens73:
I brought this up earlier today, but why do we need MSI/MESI protocol as opposed to just a simpler valid/invalid option? The point of MSI protocol is that your control is distributed among caches. We have a different use case here. Yes, data is distributed, but there's only one central controller. I feel like our states could be valid, invalid, and None.

- Valid: Most current copy of the data
- Invalid: Old copy of the data that hasn't been garbage collected so that if we need to copy data there we don't need to reallocate memory
- None: No data at all

I think this would support all the operations we need?

dialecticDolt:
> > > 
> > 
> > 
> > Agree on lock, although I think `auto_move` is called by data-movement tasks, which are scheduled to run on any thread
> 
> Isn't the movement task run in sequential? @nicelhc13

They're enqueued like other tasks in the current PR:

When task is constructed:
https://github.com/ut-parla/Parla.py/blob/2954c014b68e6841ec0e45fa23410cf269a6af62/parla/task_runtime.py#L1204

When task executes:
https://github.com/ut-parla/Parla.py/blob/2954c014b68e6841ec0e45fa23410cf269a6af62/parla/task_runtime.py#L548

nicelhc13:
Regarding NVLink, this layer is hidden from programmers. This is hardware feature, supporting more bandwidth and P2P (PCIe also can support this through GPUDirect of NVDIA).

parla/parray/coherence.py
```diff
@@ -0,0 +1,241 @@
+from __future__ import annotations # To support self references in type checking, must be first line
```
nicelhc13, line 1: Don't need this comment.

yinengy: This is for type hint when refer to itself

yinengy:
> I brought this up earlier today, but why do we need MSI/MESI protocol as opposed to just a simpler valid/invalid option? The point of MSI protocol is that your control is distributed among caches. We have a different use case here. Yes, data is distributed, but there's only one central controller. I feel like our states could be valid, invalid, and None.
> 
> * Valid: Most current copy of the data
> * Invalid: Old copy of the data that hasn't been garbage collected so that if we need to copy data there we don't need to reallocate memory
> * None: No data at all
> 
> I think this would support all the operations we need?

How do you distinguish the case between multiple valid reader and the case that only one writer is valid? You still have to add another state represent Modification. And I think the performance should be similar.

parla/parray/coherence.py
dialecticDolt: Sounds good. might want to not make it CPU bound, since device->device is usually faster, maybe `max` instead. Or possibly load from different locations assuming in SHARED state, if two auto-moves to new devices are called at once, to prevent bottlenecking. 

But these are later tiny optimizations that likely don't matter much with all our other otherheads.

sestephens73:
> How do you distinguish the case between multiple valid reader and the case that only one writer is valid?
Multiple valid readers - they all have a valid copy. One valid writer - only the writer has a valid copy. I'm not sure this answers your question but I'm also not sure I understand the question.

> What’s more, if only one device is Valid, how do you know if it is writable or not?
If only one device is Valid then that device is writeable. It's the only valid one.

> I mark GPU 0 as valid for write, and then GPU 1 want to read the data too
This is a scenario where the task on GPU 0 would mark the data as INOUT/OUT. The task on GPU1 would mark the data as IN/INOUT. The scheduler's task annotations are what tell you the ordering of tasks, not the coherence protocol here. Task 1 is scheduled after Task 0 here by necessity.

parla/parray/coherence.py
yinengy: Oh I thought PCIe is faster, looks like I am wrong.

dialecticDolt: It depends if the system has P2P or if the cpu buffer is already pinned. Most HPC workstations do have P2P now though, so it generally should be faster.

yinengy:
> > How do you distinguish the case between multiple valid reader and the case that only one writer is valid?
> 
> Multiple valid readers - they all have a valid copy. One valid writer - only the writer has a valid copy. I'm not sure this answers your question but I'm also not sure I understand the question.
> 
> > What’s more, if only one device is Valid, how do you know if it is writable or not?
> 
> If only one device is Valid then that device is writeable. It's the only valid one.
> 
> > I mark GPU 0 as valid for write, and then GPU 1 want to read the data too
> 
> This is a scenario where the task on GPU 0 would mark the data as INOUT/OUT. The task on GPU1 would mark the data as IN/INOUT. The scheduler's task annotations are what tell you the ordering of tasks, not the coherence protocol here. Task 1 is scheduled after Task 0 here by necessity.

Well, it make sense. But since the MSI is already implemented and I don’t think it bring more complexity to slow down the protocol (in most of the case, code are the same). So i will comeback later to see whether it is better to swich to valid invalid. Or maybe there exists better protocol neither MSI nor VI since I am going to implement subarray automatic data movement which may more looks like a write back cash then.

nicelhc13:
I can see a bunch of discussions but I didn't read them yet. Implementation wise, mostly fine for me.
But I have two concerns/points.
1) Comments and some naming should be clearer. Do not need to use one sentence. I prefer multiple sentences if they can make clearer. (though my codes are also bad, but you should point me out too). Please reflect or answer my questions. 
2) We already use different stream for each task; which means that data movement task is already running on different stream.
I am not sure if copying among devices works correctly. on matmul, cupy asarray or just copy worked incorreclty while data movement task moves data.

Thank you for your works.

nicelhc13:
> > The reason why I am suggesting is this PArray is core layout for Parla runtime and as it grows up with runtime, debugging would become very hard.
> 
> Yeah, I cloud add some unit test for MSI protocol later. This PR is already tested with kmeansplusplus.py so I think it may works the same as simple unit test. If there is still hidden bugs, I think it requires more complicated applications to find out.

I understand kmenas++ is complex app. But MSI protocol has multiple execution paths. I trust you but it would be the best if you have simple multiple unit tests that could handle most execution paths. (I also need this for my module though i don't know how to do it) If you need my helps, I could work on that too. 

For example,
1) def test1():
       test_data: parray
       test_data.register_dev(0)
       test_data.regsiter_dev(1)
       with cuda device 1:
           test_data.write(something)
       with cuda device 0:
           test_data.status() == invalid

blah blah.
I guess the above case is already handled by your app but you got my point.

nicelhc13:
> > @yinengy Some all of the coherence class (and maybe parts of parray) is going to need to be made threadsafe, what are the current plans for that?
> > I think it only matters for read() changing the owner at the moment.
> > But for the far future: if the plan ever changes to add versions, and reference out-of-date copies, like Mattan mentioned today: .items() can change non-safely if an IN: task registers and makes a new copy of an old-version during an update or evict execution to create a new-version, etc.
> 
> I am also thinking about threadsafe. But I just don't know how many threads will be used by parla scheduler. if there is only 1 dedicated thread for sheduler, it means the protocol info will always updated by only one thread (since only auto_move trigger states change in MSI, and auto_move is called by just scheduler)
> 
> But if threadsafe is a issues in the future. One easy solution maybe let each PArray keep a lock. When do write and read, grab the lock first.

No there is one single scheduler, but each task gets one thread or multiple threads. 
Data movement task also could get one or multiple threads.
They could cause data races if multiple tasks handle the same parray. This should be thread-safe based on our design.

nicelhc13:
> @nicelhc13 I think there is minor changes need to be made on scheduler since PArray now has new API, I think what you need is using `_auto_move` to move data. The `do_write` arguments should be set based on whether the data marked as IN or as INOUT/OUT.

Thanks. Yes. but besides this, my PR needs to be updated further. First, let's merge your code and Sean's code first. I will update auto_move part of my PR after yours are done.

nicelhc13:
> > > How do you distinguish the case between multiple valid reader and the case that only one writer is valid?
> > 
> > 
> > Multiple valid readers - they all have a valid copy. One valid writer - only the writer has a valid copy. I'm not sure this answers your question but I'm also not sure I understand the question.
> > > What’s more, if only one device is Valid, how do you know if it is writable or not?
> > 
> > 
> > If only one device is Valid then that device is writeable. It's the only valid one.
> > > I mark GPU 0 as valid for write, and then GPU 1 want to read the data too
> > 
> > 
> > This is a scenario where the task on GPU 0 would mark the data as INOUT/OUT. The task on GPU1 would mark the data as IN/INOUT. The scheduler's task annotations are what tell you the ordering of tasks, not the coherence protocol here. Task 1 is scheduled after Task 0 here by necessity.
> 
> Well, it make sense. But since the MSI is already implemented and I don’t think it bring more complexity to slow down the protocol (in most of the case, code are the same). So i will comeback later to see whether it is better to swich to valid invalid. Or maybe there exists better protocol neither MSI nor VI since I am going to implement subarray automatic data movement which may more looks like a write back cash then.

Yes I agree. If you already implemented this, just left TODO that we may simplify protocol. 
Or you can add new class of just checking validity as Sean's suggested.
So we have the highest class, sometihng like MemoryCoherencyProtocol, and we could make different policies inherited by the parent class.(But I don't read deeply/understand you guys discussion yet)

yinengy:
> Yineng, could you please specify what parts you are trying to make thread safe? Is that only for coherency protocol? or all operations on data types?

https://github.com/ut-parla/Parla.py/pull/100/commits/186ebf7bbb9ca871c59a9e384e07d9f0a8d972b5

It is this commit. There is lock around operations of updating protocol and the following data movement operations too.

nicelhc13:
> > Yineng, could you please specify what parts you are trying to make thread safe? Is that only for coherency protocol? or all operations on data types?
> 
> [186ebf7](https://github.com/ut-parla/Parla.py/commit/186ebf7bbb9ca871c59a9e384e07d9f0a8d972b5)
> 
> It is this commit. There is lock around operations of updating protocol and the following data movement operations too.

I am not sure how coherency protocol will be used. So I cannot decide if lock is required or not.
But data movement operations seem like do not need lock operations.
That should be handled by scheduler or other runtime.
Also the current implementation does not impose multi threading situations to PArray.
Data movement tasks should run if all dependent tasks are done (which moves the same or overlapped data)
Could you please any example that I am missing here?

@sestephens73 @dialecticDolt  what do you guys think about this?

yinengy:
> > > Yineng, could you please specify what parts you are trying to make thread safe? Is that only for coherency protocol? or all operations on data types?
> > 
> > 
> > [186ebf7](https://github.com/ut-parla/Parla.py/commit/186ebf7bbb9ca871c59a9e384e07d9f0a8d972b5)
> > It is this commit. There is lock around operations of updating protocol and the following data movement operations too.
> 
> I am not sure how coherency protocol will be used. So I cannot decide if lock is required or not. But data movement operations seem like do not need lock operations. That should be handled by scheduler or other runtime. The current implementation does not impose multi threading situations to PArray. Data movement tasks should run if all dependent tasks are done (which moves the same or overlapped data) Could you please any example that I am missing here?
> 
> @sestephens73 @dialecticDolt what do you guys think about this?

Scheduler should only call API to launch a data movement operation. The operation itself is handled by PArray since scheduler how no knowledge of where latest data is and where to copy it back (it is still this case when fine grained data movement is implemented since this involved lots of dirty copy). What scheduler only know is which device need the data.

Thank it as cache. Linux kernel doesn't care how cache works and which cache line is hit. In our scenario, PArray is cache, scheduler is Linux kernel.

And in regard of coherence protocol. It is invisible to scheduler and handled by PArray. Scheduler don't need to take care of the protocol itself. When two tasks are updating the same PArray (e.g. read from the same parray), it should come from INVALID to SHARED. If MSI protocol is not locked, there will be an issue since two tasks may both seen an INVALID state and both try to upgrade it to SHARED (but in sequential execution, only one should see INVALID, and the other should see SHARED).

Since each data movement is bind to a MSI protocol update. Lock the MSI alone is not enough. 
For example, T1 update it from INVALID to SHARED.  Since MSI is locked, T2 begin and seen SHARED. At this point, T1 becomes the new data owner (assume T1 is a GPU and data was on CPU). 

Now, how to move data? If the following operation is not locked. T2 begins first and try to copy data from T1 since owner is T1. But T1 haven't finished moving the data from CPU yet even if the protocol said it is the owner. Then, there is an error.

I think this explained why not only protocol but also the following movement operation is locked.

nicelhc13:
Thank you Yineng. 
Scheduler only knows where to move. But scheduler should know which tasks use the same data blocks to make dependency.
In naive way, any data movement task moving the same data block should have dependencies; never data move tasks cannot be overlapped in this case.

Then, which case we can allow data move tasks taking care of the same blocks? (this is the case when task parallelism should be  cared)

I think this is the case:
t1 -> IN = [a, b] Out = [c]
t2 -> IN = [c], ..
t3 -> IN = [c] ..

In this case t2 and t3 are depend on t1, and they could run in parallel. Right?

so to speak, tasks only overlapping IN parts are called 'shared' and could be run in parallel, and that's why locking isrequired on parray. right?

yinengy:
> Thank you Yineng. Scheduler only knows where to move. But scheduler should know which tasks use the same data blocks to make dependency. In naive way, any data movement task moving the same data block should have dependencies; never data move tasks cannot be overlapped in this case.
> 
> Then, which case we can allow data move tasks taking care of the same blocks? (this is the case when task parallelism should be cared)
> 
> I think this is the case: t1 -> IN = [a, b] Out = [c] t2 -> IN = [c], .. t3 -> IN = [c] ..
> 
> In this case t2 and t3 are depend on t1, and they could run in parallel. Right?
> 
> so to speak, tasks only overlapping IN parts are called 'shared' and could be run in parallel, and that's why locking isrequired on parray. right?

Yes, this is the same as my example. It shows the data movement operations should be locked. Actually, it means the `auto_move` is locked.

nicelhc13:
Sorry, but still confused. My question is why two tasks try to update the same data? Don't they have dependencies?

yinengy:
> Sorry, but still confused. My question is why two tasks try to update the same data? Don't they have dependencies?

It is not UPDATE, it is MOVE data. Even if the data is read only, it still need to be copied from where is was.

yinengy:
So is there any other case that scheduler will let two threads update the same PArray expect the example we mentioned?

nicelhc13:
> So is there any other case that scheduler will let two threads update the same PArray expect the example we mentioned?

That was my question. What I can think of is no. Could you please remind me what was the motivation of data coherency?
What I remember was we should track data location. (So, it requires meta information for data, like PTask) And you told me we can already track by Numpy and Cupy interfaces. And then why did we decide to implement and exploit data coherency?
I missed that discussion.

yinengy:
> > So is there any other case that scheduler will let two threads update the same PArray expect the example we mentioned?
> 
> That was my question. What I can think of is no. Could you please remind me what was the motivation of data coherency? What I remember was we should track data location. (So, it requires meta information for data, like PTask) And you told me we can already track by Numpy and Cupy interfaces. And then why did we decide to implement and exploit data coherency? I missed that discussion.

So CuPy/NumPy could handle HOW data is copied, but they don't know WHERE the data is from or to. In our system, scheduler resolve where copy to, and MSI resolve where copy from. Without a coherence protocol, every device keeps its own copy, how do you know which device's copy is updated? How does a device know its local copy is out of dated?

nicelhc13:
Cool. It makes sense to me that memory coherency is required.
But two questions are still unclear to me.
1. Is the current design choice for multi-thread handling the best or reasonable if we consider the points that I listed?
2. Will you allow one writer-multiple readers or multiple writers (My understanding is one writer, right?)?
3. What happened at this case?
   3.1. Two tasks are running at the same time, each of them a writer and a reader, respectively.
   3.2. Let's assume that all two tasks get the same status data block. (e.g. T1 and T2 get b1. In b1, 1 means the first state) 
   3.3. One task writes b1 to b2. The another task is still running with b1's copy.
   **3.4. First question, is it possible programming model? Can we run a writer and a reader at the same time? Don't they have a data dependency?**
   3.5. If it is not allowed, there is no need for parallelism, isn't it?
   3.6. **Second question, if it is allowed, you have two options: atomic and transaction ways. What is your plan?**The former needs to stop the T2, update T2's data block and resume T2. FYI, this could be expensive. This is different from cache/kernel in terms of performance since they have different hardware from CPU, memory managers. The latter just runs T1 and T2 at the same time. But T1's state is static. T1's data is still b1 and computes based on that. T2 will produce b2 and b's global status is b2.
In this case, we don't need lock. Only data of T2 will change the data, and this could be global after T1 and T2 are done by setting T1's data block as invalid, and T2's one is only valid.
   3.7. Depending on the above design choice, we actually don't need MSI protocol, but could use much simplified protocol. But still I am fine with using the MSI protocol. We can implement and append other protocols later. But we should clarify how it works, and what will need. I am sorry for continuous interrupts but please help me to understand this.

yinengy:
The number of writers and readers are defined by protocol. For MSI, if there is a writer, on other reader exists. For other protocol, idk.

nicelhc13:
I cannot understand that too. We are aiming to Parla runtime, not disjointed coherency runtime. We should have clear semantic/purpose of that coherency. Let's talk about this in Slack.

yinengy:
What I want to say it, even if you spend lots of time to discuss this rather than stick to what we currently have (MSI works, isn't it?). All this thing will not be the same in next step when fine grained data movement is required. Even if you come up with a better protocol for current application, it may not be sufficient in the future.

nicelhc13:
I think it is worth. The reason why I am trying to understand this is b/c it is related to policy and mechanism of Parla runtime. I don't argue that it is good/bad. Why are we doing this and what behaviors we expect. Otherwise, how can we handle data movement task? Let's move to Slack.

nicelhc13 (changes requested):
req

parla/parray/coherence.py
```diff
@@ -0,0 +1,241 @@
...
+    @staticmethod
+    def load(dst, src) -> MemoryOperation:
+        return MemoryOperation(MemoryOperation.LOAD, dst, src)
+
+    @staticmethod
+    def evict(src) -> MemoryOperation:
+        return MemoryOperation(MemoryOperation.EVICT, src=src)
+
+
+class Coherence:
```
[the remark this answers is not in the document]
yinengy: This is an interface, Maybe I could make a interface class and inherit it.

yinengy (commented):
Restyle as Hochan suggested.

parla/parray/coherence.py
```diff
@@ -0,0 +1,241 @@
...
+                self._overall_state = self.MODIFIED if do_write else self.SHARED
+                return operations
+
+    def read(self, operator: int) -> MemoryOperation:
+        """ Tell the protocol that operator read from the copy.
+
+        The operator should already be registered in the protocol.
+
+        Args:
+            operator: device id of the operator
```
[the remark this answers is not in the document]
nicelhc13: generally in computer science, operator means higher order functions for operations (https://en.wikipedia.org/wiki/Operator_(computer_programming)) so this is still confusing even though I now understand this. I can see the comment of this at line 67. so for now, it is fine for me. but plz keep this in your mind for later.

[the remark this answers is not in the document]
nicelhc13: nvm. you modified at the next commit.

nicelhc13 (approved):
LGTM

PR #108, opened 2022-03-18
Runahead scheduling

nicelhc13:
Tried to avoid to import Parla.cuda for cpu tasks.
also thought about extensibility for multiple devices and other architectures.


yinengy (approved):
Looks good

parla/task_runtime.py
```diff
@@ -1319,6 +1420,35 @@ def _schedule_tasks(self):
...
+            worker = self._free_worker_threads.pop() # grab a worker
+            logger.info(f"[Scheduler] Launching CPU task, %r on %r",
+                        task, worker)
+            # Assign the task to the worker (this notifies the worker's monitor)
+            worker.assign_task(task)
+            logger.debug(f"[Scheduler] Launched %r", task)
+        else:
+            queue.appendleft(task)
+
+    def _launch_gpu_tasks(self, queue, task: Task, dev: Device):
```
sestephens73, line 1441: `_launch_gpu_task`

parla/task_runtime.py
```diff
@@ -1319,6 +1420,35 @@ def _schedule_tasks(self):
...
                 self._device_queues[d].append(task)
 
+    # _launch_[DEVICE TYPES]_tasks launches a task by assigning it to available
+    # worker threads. It manages different execution paths based on the target
+    # devices' types.
+    # For example, in the future, the runtime will allow two task colocations
+    # of computation, moving data in, and moving data out.
+    # TODO(lhc): for now, the CPU/GPU launchers are same.
+
+    def _launch_cpu_tasks(self, queue, task: Task, dev: Device):
```
sestephens73, line 1430: `_launch_cpu_task`

parla/task_runtime.py
```diff
@@ -515,20 +536,65 @@ def run(self):
...
-                    ctx.scheduler._available_resources.deallocate_resources(d, self.req.resources)
+                    ctx.scheduler._available_resources. \
+                              deallocate_resources(d, self.req.resources)
+                # Regardless of the previous notification,
+                # (So, before leaving the current run(), the above)
+                # it should notify dependees since
+                # new dependees could be added after the above
+                # notifications, while other devices are running
+                # their kernels asynchronously.
+                self._notify_dependees_mutex()
```
sestephens73, line 597: I'm concerned about a potential race condition here where a task spawns after this notify but before the task state is set to finished... Is there a way to do this and the next line atomically?

sestephens73 (approved):

parla/task_runtime.py
nicelhc13: Or how about merging the two functions within one mutex statement?
For example, 
``` 
with self.mutex:
     self._notify_dependees()
     self._set_state(task_state)
```

Could it be reasonable?

sestephens73: IIUC, the mutex doesn't make it atomic, it just protects the data structures. If dependees checking the task's state must acquire that same mutex to access the state, I think this will work.

nicelhc13: Yes, generally this is atomic vs transaction. I just suggested this because I have no clue on Python atomic :-) It should share the data strucutre iiuc. So if you think it is fine, we are on board. Let me add this and merge it. Thank you for ur suggestion

PR #109, opened 2022-03-21
Loosen restraint to allow concurrent data movement tasks if parent tasks have dependency

dialecticDolt:
task

^ removing red edges in this picture, to allow blue edges to happen concurrently. 

This makes all data movement tasks reads only. 
Write invalidation is moved to when a compute_task starts. 

Note atm.... Parrays still have a global lock, so concurrent reads can't happen in practice. This should be fixed in the same PR slicing is added.


dialecticDolt:
Closing this to create a new PR because this became quickly too outdated to be maintainable.

PR #110, opened 2022-03-21
Mapper memory tracking and policy updates

sestephens73:
- Added new mapper algorithm to find device with best suitability based on load, local data, and dependency location
- Got the new mapper algorithm to properly assign devices to tasks
   - Mapper creates a new TaskEnvironment with chosen device, just like in the old system
- Updated resource allocation/deallocation calls to be consistent throughout task_runtime.py
   - All resources are allocated within the map phase after placements are chosen
   - All resources are deallocated on task cleanup
   - VCUs have no effect on GPU devices at the moment
- Updated parray tracking in task_runtime.py
   - At the end of a task, all of the task's OUT/INOUT parrays have their tracking updated in case their size changed or their coherence updated
- Modified device affinity for tasks with dependencies
   - If you have a dependency task running on one device AND more data on that device than elsewhere, always run there regardless of load
- Tested the above changes on synthetic serial graph, synthetic independent graph, matmul_automove, and blocked_cholesky_parray

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -1,3 +1,4 @@
+# General imports
 from functools import lru_cache
 import logging
 import random
@@ -10,19 +11,17 @@ import time
 from itertools import combinations
 from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSet, Iterable, TypeVar, Deque
 
+# Parla imports
 from parla.device import get_all_devices, Device
 from parla.environments import TaskEnvironmentRegistry, TaskEnvironment
-
 from parla.cpu_impl import cpu
 
+# Logger configuration (uncomment and adjust level if needed)
 #logging.basicConfig(level = logging.INFO)
 logger = logging.getLogger(__name__)
 
+
 __all__ = ["Task", "SchedulerContext", "DeviceSetRequirements", "OptionsRequirements", "ResourceRequirements", "get_current_devices"]
-
-# TODO: This module is pretty massively over-engineered the actual use case could use a much simpler scheduler.
-
-_ASSIGNMENT_FAILURE_WARNING_LIMIT = 32
 
 
 # Note: tasks can be implemented as lock free, however, atomics aren't really a thing in Python, so instead
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -15,7 +15,6 @@ from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSe
 from parla.device import get_all_devices, Device
 from parla.environments import TaskEnvironmentRegistry, TaskEnvironment
 from parla.cpu_impl import cpu
-from parla.cuda import gpu
 
 # Logger configuration (uncomment and adjust level if needed)
 #logging.basicConfig(level = logging.INFO)
@@ -986,7 +985,7 @@ class ResourcePool:
                 dres = self._devices[d]
 
                 # Workaround stupid vcus (I'm getting rid of these at some point)
-                if d.architecture == gpu and name == 'vcus':
+                if d.architecture.id == 'gpu' and name == 'vcus':
                     continue
 
                 if amount > dres[name]:
@@ -1021,7 +1020,7 @@ class ResourcePool:
 
     def _update_resource(self, dev: Device, res: str, amount: float, block: bool):
         # Workaround stupid vcus (I'm getting rid of these at some point)
-        if dev.architecture == gpu and res == 'vcus':
+        if dev.architecture.id == 'gpu' and res == 'vcus':
             return True
         try:
             while True: # contains return
@@ -1049,7 +1048,7 @@ class ResourcePool:
     def _to_parray_index(self, device):
         if device.architecture == cpu:
             return self.CPU_INDEX
-        if device.architecture == gpu:
+        if device.architecture.id == 'gpu':
             return device.index
         raise NotImplementedError("Only cpu and gpu architectures are supported")
 
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -15,6 +15,7 @@ from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSe
 from parla.device import get_all_devices, Device
 from parla.environments import TaskEnvironmentRegistry, TaskEnvironment
 from parla.cpu_impl import cpu
+from parla.cuda import gpu
 
 # Logger configuration (uncomment and adjust level if needed)
 #logging.basicConfig(level = logging.INFO)
@@ -177,6 +178,16 @@ ResourceDict = Dict[str, Union[float, int]]
 
 
 class ResourceRequirements(object, metaclass=ABCMeta):
+    """
+    When a task spawns, it has a set of requirements based on parameters
+    supplied to @spawn.
+    This class represents those resources.
+    This is an Abstract Base Class - see below for classes which inherit from it.
+    Currently, spawned tasks only use DeviceSetRequirements.
+    After mapping, tasks receive EnvironmentRequirements.
+    As of writing this comment, idk what the difference is. Enviroments seem unnecessary and confusing.
+    OptionsRequirements aren't even used anywhere at all.
+    """
     __slots__ = ["resources", "ndevices", "tags"]
 
     tags: FrozenSet[Any]
@@ -226,6 +237,8 @@ class EnvironmentRequirements(ResourceRequirements):
         return "EnvironmentRequirements({}, {})".format(self.resources, self.environment)
 
 
+# This basically stores all the devices a task is *permitted* to run on,
+# taking into account spawn's placement parameter
 class DeviceSetRequirements(ResourceRequirements):
     __slots__ = ["devices"]
     devices: FrozenSet[Device]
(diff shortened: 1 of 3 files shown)


parla/task_runtime.py
```diff
@@ -141,7 +141,8 @@ def clear_dependencies(self):
 
     def __repr__(self):
         if self.func:
-            return "TaskRunning({}, {}, {})".format(self.func.__name__, self.args, self.dependencies)
```
nicelhc13, line 144: Why should you remove this?

parla/task_runtime.py
```diff
@@ -178,6 +179,16 @@ def __repr__(self):
...
+    """
+    When a task spawns, it has a set of requirements based on parameters
+    supplied to @spawn.
+    This class represents those resources.
+    This is an Abstract Base Class - see below for classes which inherit from it.
+    Currently, spawned tasks only use DeviceSetRequirements.
+    After mapping, tasks receive EnvironmentRequirements.
+    As of writing this comment, idk what the difference is. Enviroments seem unnecessary and confusing.
+    OptionsRequirements aren't even used anywhere at all.
+    """
```
nicelhc13, line 191: This is good! This is what we need!

parla/task_runtime.py
```diff
@@ -414,7 +433,7 @@ def _complete_dependency(self):
 
     def _set_state(self, new_state: TaskState):
         # old_state = self._state
-        logger.info("Task %r: %r -> %r", self, self._state, new_state)
+        logger.info("[Task] %r: %r -> %r", self._name, self._state, new_state)
```
nicelhc13, line 436: I noticed that generally str(self.taskid) is more useful than self._name. But it's fine for now.

parla/task_runtime.py
```diff
@@ -595,10 +617,16 @@ def run(self):
...
-                    ctx.scheduler._available_resources.deallocate_resources(d, self.req.resources)
+
+                # DON'T deallocate resources!
+                # DataMovementTask has the same resources as the ComputeTask which created it
+                # That ComputeTask will do the deallocation
+                # If we do it here, we overly deallocate
+
+                # Don't parray tracking information either
+                # The scheduler already registered the new location
+                # If size changes, the ComputeTask will take care of that
```
nicelhc13, line 628: gooood

parla/task_runtime.py
```diff
@@ -936,8 +984,14 @@ def check_resources_availability(self, d: Device, resources: ResourceDict):
             is_available = True
             for name, amount in resources.items():
                 dres = self._devices[d]
+
+                # Workaround stupid vcus (I'm getting rid of these at some point)
```
nicelhc13, line 988: Does workaround mean it just runs with no effects?

parla/task_runtime.py
```diff
@@ -10,19 +11,18 @@
 from itertools import combinations
 from typing import Optional, Collection, Union, Dict, List, Any, Tuple, FrozenSet, Iterable, TypeVar, Deque
 
+# Parla imports
 from parla.device import get_all_devices, Device
 from parla.environments import TaskEnvironmentRegistry, TaskEnvironment
-
 from parla.cpu_impl import cpu
+from parla.cuda import gpu
```
nicelhc13, line 18: I missed this discussion. What was plan? Just allow to import gpu  (with assumption that our targets must have gpu tasks?)

If it is, could you please add TODO or NOTE that it can cause performance overhead?

parla/task_runtime.py
```diff
@@ -985,6 +1041,87 @@ def _update_resource(self, dev: Device, res: str, amount: float, block: bool):
         except KeyError:
             raise ValueError("Resource {}.{} does not exist".format(dev, res))
 
+    ### PARRAY MEMORY TRACKING CALLS ###
```
nicelhc13, line 1044: Just curious. Can these functions be in the parray/core.py later?

parla/task_runtime.py
```diff
@@ -1161,6 +1302,106 @@ def _assignment_policy(self, task: Task):
...
+            print(f"nonlocal={myformat(nonlocal_data)}", end='   ')
+            print(f"load={myformat(dev_load)}", end='   ')
+            print(f"suit={myformat(suitability)}")
+            """
+
+            # Update whether or not this is the most suitable device
+            if max_suitability is None or suitability > max_suitability:
+                max_suitability = suitability
+                best_device = device
+                best_device_owns_dependency = this_device_owns_dependency
```
nicelhc13, line 1387: Could you plesae describe 'own' meaning in this context?

parla/task_runtime.py
dialecticDolt: Seconded!

parla/task_runtime.py
```diff
@@ -282,14 +296,19 @@ def __init__(self, dependencies: Collection["Task"], taskid,
                  req: ResourceRequirements, name: Optional[str] = None):
         self._mutex = threading.Lock()
         with self._mutex:
+            # This is the name of the task, which is distinct from the TaskID and from the name of its func?
```
dialecticDolt, line 299: It inherits the name of its function in `@spawn` iirc.

parla/task_runtime.py
dialecticDolt: Seconding this too. I don't think the names are unique generally.

dialecticDolt: Either is better than just dumping the whole object like before though

parla/tasks.py
```diff
@@ -418,7 +418,7 @@ def decorator(body):
         taskid.dependencies = dependencies
 
         # gather input/output/inout, which is hint for data from or to the this task
-        dataflow = Dataflow(input, output, inout)
+        dataflow = Dataflow(list(input), list(output), list(inout))
```
dialecticDolt, line 421: Why was this added? 

Also side comment for the future, I think we should add a flatten for input/output/inout in `spawn`. Just in case the user passes a list of lists, etc. (2-matrix of parrays)

parla/task_runtime.py
```diff
@@ -525,10 +540,20 @@ def run(self):
...
+                    for resource, amount in self.req.resources.items():
+                        logger.debug("Task %r deallocating %d %s from device %r", self, amount, resource, d)
                     ctx.scheduler._available_resources.deallocate_resources(d, self.req.resources)
+                    ctx.scheduler._device_compute_task_counts[d] -= 1
+
+                # Update parray tracking information
+                # The easiest way to do this is just untrack and re-track the array for anything that could've changed
+                for parray in (self.dataflow.inout + self.dataflow.output):
+                    ctx.scheduler._available_resources.untrack_parray(parray)
+                    ctx.scheduler._available_resources.track_parray(parray)
```
dialecticDolt, line 555: Would there be any problems if the GIL switches to another thread/task between 554 and 555?

dialecticDolt (approved):
lgtm. Tracking and policy might need some changes for eviction, but thats def out of scope for the deadline rn

parla/tasks.py
dialecticDolt: disregard second part. Its fine if this is on the user, its too expensive to put in every spawn calll.

parla/task_runtime.py
sestephens73: It printed so much information that it made debugging too messy for me haha. I've left the original version commented there in case anyone needs it later. If you'd like, I can revert it.

parla/task_runtime.py
sestephens73: Yup. For now at least. None of our apps use these at the moment. I think we should worry about fixing them later, but not this week.

parla/task_runtime.py
sestephens73: The device "owns" the dependency task, or in other words, the dependency task is mapped to the device.

parla/tasks.py
sestephens73: I wanted to iterate through `input + inout` or `output + inout`, etc., in various places, but tuples can't be concatenated, so I just moved them to lists.

I suppose that could cause a perf hit... Do you think I should just go rework all of those into separate `for` loops?

parla/task_runtime.py
nicelhc13: no problem. it's fine for me.

parla/task_runtime.py
nicelhc13: unfortunately, there was no clear solution. our solution for parray and event was just to hide importing if we don't need. if we don't have clear answer, plz just leave TODO to revisit this.

FYI, you can see performance differences through quicksort. 
``` python examples/quicksort.py```

parla/task_runtime.py
nicelhc13: Then, it looks like putting this on scheduler looks better. ok thanks.

nicelhc13 (approved):
LGTM

parla/tasks.py
dialecticDolt: Maybe refactor to do the concatenation once at dataflow __init__ ?

dialecticDolt: I don't think its the major source of overhead right now, we've got a death by a thousand slightly larger cuts first.

parla/task_runtime.py
sestephens73: aaaah this is making me realize i have several mutexes where GIL switching could mess things up

sestephens73: I'm adding it as a TODO. I think it's unlikely to bite us rn (knock on wood) and it's something I can look at fixing later when I'm also fixing evictions to free memory.

parla/task_runtime.py
sestephens73: I've fixed this. The fix is simple. When checking for GPU architecture, instead of

`if architecture == gpu`

just do

`if architecture.id == 'gpu'`

This requires no import.

parla/tasks.py
sestephens73: Added as a TODO comment for later

PR #111, opened 2022-03-24
Fix data race among multiple readers

nicelhc13:


yinengy (approved):
This is good. Thanks.

PR #114, opened 2022-03-24
Attempt to fix intialization crash by making initialization seqential

dialecticDolt:
This PR adds a lock around thread initialization, and device synchronizations between every handle creation. 

This is absolutely overkill and will slow down initialization. 

But hopefully this resolves the crash we're seeing in the short term.


sestephens73 (approved):

nicelhc13 (approved):
LGTM

PR #115, opened 2022-03-25
Mapper dependency load balance

sestephens73:
- Added a random mapping option
- Previously I tried to make any dependency run on the same device as its parent task if its data is on that device. This leads to terrible load balancing if the parent had multiple dependees. Reworked this to be a weighted constant.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -511,6 +511,7 @@ class ComputeTask(Task):
                 self.notify_unspawned_dependees()
                 self._state = TaskRunning(func, args, None)
                 get_scheduler_context().incr_active_tasks()
+                get_scheduler_context().scheduler.incr_active_compute_tasks()
                 # Enqueue this task right after spawning on the spawend queue.
                 # The task could have dependencies.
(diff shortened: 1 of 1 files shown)


parla/task_runtime.py
```diff
@@ -1490,52 +1503,50 @@ def _assignment_policy(self, task: Task):
...
             nonlocal_data_weight = 10.0
             load_weight = 1.0
+            dependency_weight = 0.5
 
             # Calculate the suitability
             suitability = local_data_weight * local_data \
                         - nonlocal_data_weight * nonlocal_data \
-                        - load_weight * dev_load
+                        - load_weight * dev_load \
+                        + dependency_weight * this_device_owns_dependency
```
dialecticDolt, line 1532: just a future note I think this should depend on the amount of data shared with the dependency maybe

dialecticDolt (approved):
lgtm. I have no idea what the right constant should be to prevent unbalance, but this seems like it'll help

nicelhc13 (approved):
LGTM

PR #116, opened 2022-03-26
Organize concurrent read dependencies

dialecticDolt:
This PR is the same as the old one. 

Changes (most of which are implemented in a really lazy way):
- Merge in CV Parray changes
- Compute tasks that use a data block as IN are never used as a dependee
- OUT data movement tasks are removed (since this is just INOUT)
- All data movement tasks now only perform read operations
- Write operations are performed at the start of compute tasks. 
---- In the current PR, this is done with auto_move(). Optimally movement should never occur. 

 Note: This PR is the same as the old one. 

Changes (most of which are implemented in a really lazy way):
- Merge in CV Parray changes
- Compute tasks that use a data block as IN are never used as a dependee
- OUT data movement tasks are removed (since this is just INOUT)
- All data movement tasks now only perform read operations
- Write operations are performed at the start of compute tasks. 
---- In the current PR, this is done with auto_move(). Optimally movement should never occur. 

 Note: 
**I think this has the same hanging bug as current dev. Making this PR because I'm not sure its a new different cause, review with caution!**

task

^ removing red edges in this picture, to allow blue edges to happen concurrently.

### DIFF
--- a/examples/cholesky/blocked_cholesky_gpu.py
+++ b/examples/cholesky/blocked_cholesky_gpu.py
@@ -60,6 +60,7 @@ num_tests = args.trials
 
 loc = gpu
 
+save_file = True
 check_nan = False
 check_error = True
 
@@ -262,6 +263,7 @@ def cholesky_blocked_inplace(a):
 def main():
     @spawn(placement=cpu)
     async def test_blocked_cholesky():
+        global n
 
         if args.matrix is None:
             print("Generating matrix of size: ", n)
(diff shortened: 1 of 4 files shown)


dialecticDolt:
ah ignore this. Forgot to push some changes!

dialecticDolt:
Now its updated.

yinengy:
condition variable changes are included in #119

dialecticDolt:
Closing this because my other PR includes this. Whoops.

PR #118, opened 2022-03-27
Merge of all things related to data movement optimization

dialecticDolt:
Still being tested. 

But here for review.

### DIFF
--- a/parla/parray/coherence.py
+++ b/parla/parray/coherence.py
@@ -130,7 +130,7 @@ class Coherence:
             else:
                 return MemoryOperation.check_data(device_id)  # check if the data is ready
 
-    def write(self, device_id: int) -> List[MemoryOperation]:
+    def write(self, device_id: int, copy: bool=True) -> List[MemoryOperation]:
         """ Tell the protocol that this device write to the copy.
 
         Args:
@@ -150,7 +150,10 @@ class Coherence:
                     prev_owner = self._owner
 
                     self._data_ready[device_id] = False
-                    operations = [MemoryOperation.load(dst=device_id, src=prev_owner)]
+                    if copy:
+                        operations = [MemoryOperation.load(dst=device_id, src=prev_owner)]
+                    else:
+                        operations = [MemoryOperation.noop()]
 
                     # evict data from other devices
                     for device, state in self._local_states.items():
@@ -168,7 +171,11 @@ class Coherence:
                     # load data from previous owner
                     prev_owner = self._owner
                     self._data_ready[device_id] = False
-                    operations = [MemoryOperation.load(dst=device_id, src=prev_owner)]
+
+                    if copy:
+                        operations = [MemoryOperation.load(dst=device_id, src=prev_owner)]
+                    else:
+                        operations = [MemoryOperation.noop()]
 
                     # evict data from previous owner
                     self._local_states[prev_owner] = self.INVALID
--- a/parla/parray/coherence.py
+++ b/parla/parray/coherence.py
@@ -154,6 +154,7 @@ class Coherence:
                         operations = [MemoryOperation.load(dst=device_id, src=prev_owner)]
                     else:
                         operations = [MemoryOperation.noop()]
+                        self._data_ready[device_id] = True
 
                     # evict data from other devices
                     for device, state in self._local_states.items():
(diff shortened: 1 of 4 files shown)


parla/task_runtime.py
```diff
@@ -594,6 +594,30 @@ def run(self):
...
+                    dev_no = -1
+                    if (dev_type.architecture is not cpu):
+                        dev_no = dev_type.index
+
+                    #Note: This should NEVER perform a copy. Only invalidate the data.
+                    #Note Note: If it does ever perform a copy, its probably fine but slow?
+                    #TODO: Add new parray syntax to disable copy and only invalidate (useful for OUTs)
+                    for data in self.dataflow.output:
+                        #print("Doing OUT invalidation for ", data._name, flush=True)
+                        data._auto_move(device_id = dev_no, do_write = write_flag, do_copy=False)
```
nicelhc13, line 614: So this does only invalidate data..

parla/task_runtime.py
```diff
@@ -1686,8 +1746,12 @@ def _construct_datamove_task(self, target_data, compute_task: ComputeTask, opera
                 if compute_task.is_dependent(dep_task):
                     if not datamove_task._add_dependency(dep_task):
                         completed_tasks.append(dep_task_id)
-            dep_task_list = [tuple(dt for dt in dep_task_list if dt[0] != ft) for ft in completed_tasks]
```
nicelhc13, line 1689: Why do you remove this line?? It is view of dictionary and removes completed tasks.

parla/task_runtime.py
```diff
@@ -594,6 +594,30 @@ def run(self):
...
+
+                    #Note: This should NEVER perform a copy. Only invalidate the data.
+                    #Note Note: If it does ever perform a copy, its probably fine but slow?
+                    #TODO: Add new parray syntax to disable copy and only invalidate (useful for OUTs)
+                    for data in self.dataflow.output:
+                        #print("Doing OUT invalidation for ", data._name, flush=True)
+                        data._auto_move(device_id = dev_no, do_write = write_flag, do_copy=False)
+                        #print("Finished OUT invalidation for ", data._name, flush=True)
+
+                    for data in self.dataflow.inout:
```
nicelhc13, line 617: inout has a datamovement task, but why are you invalidating at here again?

parla/parray/coherence.py
```diff
@@ -98,30 +100,37 @@ def read(self, device_id: int) -> MemoryOperation:
...
+                    self._data_ready[device_id] = False  # copy in progress
+                    return MemoryOperation.load(dst=device_id, src=self._owner)
+                elif self._global_state == self.MODIFIED:
+                    prev_owner = self._owner
+                    self._local_states[prev_owner] = self.SHARED
+                    self._local_states[device_id] = self.SHARED
+                    self._global_state = self.SHARED
+
+                    # Trick: smaller one becomes owner, so will always load from CPU (-1) when possible
+                    self._owner = max(self._owner, device_id)
```
nicelhc13, line 123: this is max. so comment is wrong.

parla/task_runtime.py
dialecticDolt: Because inout no longer invalidates during its datamovement task. I delayed invalidation to this step. 
In this PR data movement only performs `reads` and all `write` invalidation are done right before the data is about to be modified.

nicelhc13 (approved):
LGTM; honestly not that focused on parray part.

parla/parray/coherence.py
dialecticDolt: Yep. This is a holdover from the old parray code. I didn't touch this but will update

parla/task_runtime.py
dialecticDolt: Yep, only invalidates data! It passes a noop to the parray routines for data movement when the flag is false.
This is likely not the lightest weight way to do this. But had the least code modifications.

parla/task_runtime.py
dialecticDolt: I didn't know what it was for! It seemed like an expensive loop that make something that was never used later. 

I don't think its a view. I think the assignment creates a new list and this new list is never used to update the dictionary.

yinengy:
These commits are included in fine grained parray, we could just merge that one.

sestephens73:
What's the status of this PR? Is it outdated? Should I review it? Why are there so many merge conflicts?

dialecticDolt:
Closing this as outdated.

PR #120, opened 2022-03-27
Memory tracking fixes

sestephens73:
Previously, the scheduler didn't start tracking a parray until the parray was directly used by a task. This meant that if a parray was initialized from a cupy array on a device, memory would be used that the scheduler was unaware of. I added a callback which starts parray tracking in the parray init function.

Previously, the mapper's assignment policy mistakenly only looked at the `memory` parameter to see if devices had enough resources to run a task. Now it looks at the `memory` parameter plus the amount of data needing to be moved to the device for gather necessary parrays.

### DIFF
--- a/parla/parray/core.py
+++ b/parla/parray/core.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from parla.cpu_impl import cpu
-from parla.task_runtime import get_current_devices
+from parla.task_runtime import get_current_devices, get_scheduler_context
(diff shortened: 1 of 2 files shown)


nicelhc13 (changes requested):
LGTM

nicelhc13 (approved):
LGTM

dialecticDolt (approved):
LGTM, I think I saw most of these changes on the call.

PR #122, opened 2022-04-18
Syncing dev with main

sestephens73:
Merging Will's changes back into dev so it's up to date


dialecticDolt (approved):
Thanks!

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "examples/nbody"]
 	path = examples/nbody
-	url = git@github.com:ut-parla/barnes-hut.git
+	url = https://github.com/ut-parla/barnes-hut.git
 [submodule "examples/blr"]
 	path = examples/blr
-	url = git@github.com:ut-parla/BLR.git
```
nicelhc13, line 6: Just curious. Why is this change required?
Technically, both look fine and private repos enforces to use git@github.com format.

PR #123, opened 2022-04-28
Use per-device locks to update resource status

nicelhc13:
This PR uses per-device locks to update resource status.
This PR requires CPU core modes to see performance improvements; do `git checkout cpu-test -- parla/cpu_impl.py`.
It improved 3x faster runtime compared to the main.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -1074,14 +1074,12 @@ class ResourcePool:
 
     # Resource pools track device resources. Environments are a separate issue and are not tracked here. Instead,
     # tasks will consume resources based on their devices even though those devices are bundled into an environment.
-
     def __init__(self):
         self._monitor = threading.Condition(threading.Lock())
-
         # Devices are stored in a dict keyed by the device.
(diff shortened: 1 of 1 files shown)


parla/task_runtime.py
```diff
@@ -1074,22 +1074,28 @@ class ResourcePool:
...
+        self._device_locks = [threading.Lock() for d in range(len(self._devices))]
         # Parla tracks managed PArrays' locations
         # Index into dict with id(array), then with device. True means the array is present there
         # We use the unique id of the array as the key because PArray is an unhashable class
         self._managed_parrays = {}
 
     @staticmethod
     def _initial_resources():
-        return {dev: {name: amt for name, amt in dev.resources.items()} for dev in get_all_devices()}
+        res_dict = {}
```
dialecticDolt, line 1090: Why this change?

nicelhc13: just to give index; individual lock exists for each device. But I have no idea how to write it in more pythonic way :-)

PR #112, opened 2022-03-24
Miscellaneous refactoring and robustness enhancement

bozhiyou:
- Use `predecessor/successor` in place of ambiguous `dependee` to express dependencies
- Refactor dependency management to avoid counters +/- around 
- Group repetitive functionalities
- Fix type hints with reference forward
- Fix imports of `task_runtime` and `data flow`

PR #124, opened 2022-04-28
Miscellaneous refactoring, continued

sestephens73:
I (Sean) had to merge #112 early because it was blocking me. This is a continuation of #112 addressing comments and concerns there.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -32,11 +32,11 @@ TaskAwaitTasks = namedtuple("AwaitTasks", ("dependencies", "value_task"))
 
 
 class UnspawnedDependencies:
-    """ Collection of dependencies where the upstream tasks ("predecessors") are not spawned yet
-    and thus making downstream tasks ("successors") wait.
+    """ Collection of dependencies where the upstream tasks ("dependencies") are not spawned yet
+    and thus making downstream tasks ("dependents") wait.
     """
 
-    # A dictionary with predecessors as keys and (lists of) successors as values,
+    # A dictionary with dependencies as keys and (lists of) dependents as values,
     # i.e. value tasks are waiting for key tasks.
     #
     # Example:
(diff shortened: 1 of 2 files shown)


parla/task_runtime.py
```diff
@@ -428,84 +422,84 @@ def is_assigned(self):
...
-    def predecessors(self, predecessors: Collection['Task']):
-        self._predecessors: List[Task] = list(predecessors)
-        self._num_blocking_predecessors: int = len(predecessors)
-        for dep in predecessors:
+    def dependencies(self) -> Tuple["Task"]:
+        return tuple(self._dependencies)
+
+    @dependencies.setter
+    def dependencies(self, dependencies: Collection['Task']):
+        self._dependencies: List[Task] = list(dependencies)
```
dialecticDolt, line 430: Why is dependencies a collection and not just a list from the start?

parla/task_runtime.py
```diff
@@ -428,84 +422,84 @@ def is_assigned(self):
...
         with self._mutex:
-            self._num_blocking_predecessors -= 1
+            self._num_blocking_dependencies -= 1
             # Add events from one dependent task.
             # (We are aiming to multiple device tasks, and it would
             #  be possible to have multiple events)
             if events is not None:
-                self._predecessor_events.append(events)
+                self._dependency_events.append(events)
             self._maybe_ready_to_schedule()
```
dialecticDolt, line 500: There was a comment on the old PR about this needed a better name, just putting it here again.

sestephens73:
All the (major) comments from #112 should be addressed at this point.

Tested on matmul: `python matmul_automatic.py 1`

parla/task_runtime.py
nicelhc13: Same question; this exists from 2yr olds.

dialecticDolt: Also less importantly this type hint is not strictly accurate, as this Collection can also have taskIDs (as mentioned in the comment near there)

nicelhc13 (approved):
lgtm

PR #125, opened 2022-05-03
Update resource managing

nicelhc13:
This PR updates two main things:
1) replace lock-based resource tracking with lock-free one by making the scheduler manage.
2) move resource updates from mapping to launching; it blocked task scheduling critically.

With this update, on 2000 independent tasks + with GIL + 76000 microsec for each task + depth 100 (5MB),
it took 7 sec for each iteration.

Without this, it took more than 130 sec for each iteration.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -343,6 +343,10 @@ class Task:
         return tuple(self._dependees)
 
     @property
+    def state(self):
+        return self._state
+
+    @property
     def result(self):
         if isinstance(self._state, TaskCompleted):
             return self._state.ret
@@ -613,15 +617,8 @@ class ComputeTask(Task):
                 for d in self.req.devices:
                     for resource, amount in self.req.resources.items():
                         logger.debug("Task %r deallocating %d %s from device %r", self, amount, resource, d)
-                    ctx.scheduler._available_resources.deallocate_resources(d, self.req.resources)
                     ctx.scheduler.update_mapped_task_count(self, d, -1)
                     ctx.scheduler.update_launched_task_count_mutex(self, d, -1)
-
-                # Update OUT parrays which may have changed size from 0 to something
-                # We assume all IN and INOUT params don't change size
-                for parray in (self.dataflow.output):
-                    ctx.scheduler._available_resources.update_parray_nbytes(parray, self.req.devices)
-
                 ctx.scheduler.decr_active_compute_tasks()
 
                 # Protect the case that it notifies dependees and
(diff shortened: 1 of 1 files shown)


dialecticDolt:
Main concern is that I think @sestephens73 had a bunch of reasons for moving resource allocation for mapping. Want to make sure we can resolve those before moving it back.

nicelhc13:
yes please point me out. regardless of that this pr works, the main bottleneck was that updating resources at mapping is bad.

nicelhc13:
let me also add @yinengy for the review.

parla/task_runtime.py
```diff
@@ -1885,9 +1885,23 @@ def _schedule_tasks(self):
...
+            for device in task.req.devices:
+                for resource, amount in task.req.resources.items():
+                    logger.debug("Task %r allocating %d %s on device %r", task, amount, resource, device)
+                self._available_resources.allocate_resources(device, task.req.resources, blocking=True)
+            # Update parray tracking and task count on the device
+            for parray in (task.dataflow.input + task.dataflow.inout + task.dataflow.output):
+                if len(task.req.environment.placement) > 1:
+                    raise NotImplementedError("Multidevice not supported")
+                for device in task.req.environment.placement:
+                    self._available_resources.register_parray_move(parray, device)
```
dialecticDolt, line 1898: This seems wrong. The data movement tasks already exist and are already run before this code executes. This is only updating the information once the "parent" task is starting to be run.

dialecticDolt: Wait it also happens at creation, I'm not sure then. Why is it registered twice onto the same device?

nicelhc13: I think this should be a mistake. Then is it fine just to remove this registeration at here?

parla/task_runtime.py
```diff
@@ -1804,6 +1797,19 @@ def _map_tasks(self):
             task: Optional[Task] = self._dequeue_spawned_task()
             if task:
                 if not task.assigned:
+                    _new_launched_task = []
+                    for n_task in self._launched_tasks:
```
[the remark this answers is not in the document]
nicelhc13: Agree, this should be also refined. but the point of this PR is that we should update resource at launching, not mapping.

parla/task_runtime.py
```diff
@@ -1804,6 +1797,19 @@ def _map_tasks(self):
...
+                        task_state = n_task.state
+                        if isinstance(task_state, TaskCompleted):
+                            for dev in n_task.req.devices:
+                                self._available_resources.deallocate_resources(dev, n_task.req.resources)
```
dialecticDolt, line 1805: I think that this doesn't handle Data Movement Tasks? Their resources are still bundled with their "parents", but their resources are no longer allocated 'ahead' of time. Only compute tasks are being tracked and deallocated. This means that they currently are considered to take 'no resources' and can be over-scheduled.

dialecticDolt: Although (sorry for the stream of consciousness comments on this), doesn't this mean it can still deadlock in the same way as the current mapper? Since it can allocate a bunch of data movement tasks out of order and prevent the next dependency from being mapped due to being out of resources?

nicelhc13: Yes. still deadlock happens. Ok, let's consider this PR as the immediate solution for the performance. I think your pointing out could not be resolved by this PR. Let's consider that issue on the separate PR with more discussion.

parla/task_runtime.py
```diff
@@ -1126,48 +1123,45 @@ def check_resources_availability(self, d: Device, resources: ResourceDict):
...
-                    success = False
-                    break
-                else:
-                    to_release.append((name, v))
+        to_release = []
+        success = True
+        for name, v in resources.items():
+            if not self._update_resource(d, name, v * multiplier, block):
+                success = False
+                break
```
sestephens73, line 1147: If this function no longer uses the monitor, it should not be named "atomically"

parla/task_runtime.py
```diff
@@ -1126,48 +1123,45 @@ def check_resources_availability(self, d: Device, resources: ResourceDict):
         :param resources: The resources to deallocate.
         """
         logger.debug("[ResourcePool] Acquiring monitor in check_resources_availability()")
-        with self._monitor:
```
sestephens73, line 1129: Are we *sure* removing this monitor is safe?

parla/task_runtime.py
```diff
@@ -1126,48 +1123,45 @@ def check_resources_availability(self, d: Device, resources: ResourceDict):
         :param resources: The resources to deallocate.
         """
         logger.debug("[ResourcePool] Acquiring monitor in check_resources_availability()")
```
sestephens73, line 1125: Delete this line too if deleting the monitor

parla/task_runtime.py
```diff
@@ -1126,48 +1123,45 @@ def check_resources_availability(self, d: Device, resources: ResourceDict):
...
+            #    continue
 
-                # Workaround stupid vcus (I'm getting rid of these at some point)
-                #if d.architecture.id == 'gpu' and name == 'vcus':
-                #    continue
-
-                if amount > dres[name]:
-                    is_available = False
-                logger.debug("Resource check for %d %s on device %r: %s", amount, name, d, "Passed" if is_available else "Failed")
-            logger.debug("[ResourcePool] Releasing monitor in check_resources_availability()")
```
sestephens73, line 1142: delete this too if removing the monitor

parla/task_runtime.py
```diff
@@ -1126,48 +1123,45 @@ def check_resources_availability(self, d: Device, resources: ResourceDict):
...
-            return is_available
+            if amount > dres[name]:
+                is_available = False
+            logger.debug("Resource check for %d %s on device %r: %s", amount, name, d, "Passed" if is_available else "Failed")
+        logger.debug("[ResourcePool] Releasing monitor in check_resources_availability()")
+        return is_available
 
     def _atomically_update_resources(self, d: Device, resources: ResourceDict, multiplier, block: bool):
         logger.debug("[ResourcePool] Acquiring monitor in atomically_update_resources()")
-        with self._monitor:
```
sestephens73, line 1147: same question, are we sure this is safe to remove? Is it because only the scheduler calls this now?

parla/task_runtime.py
sestephens73: I suppose it is if *only* the scheduling thread makes this call.

parla/task_runtime.py
```diff
@@ -1179,18 +1173,14 @@ def _update_resource(self, dev: Device, res: str, amount: float, block: bool):
                 dres = self._devices[dev]
                 if -amount <= dres[res]:
                     dres[res] += amount
-                    if amount > 0:
-                        self._monitor.notify_all()
-                    assert dres[res] <= dev.resources[res], "{}.{} was over deallocated".format(dev, res)
-                    assert dres[res] >= 0, "{}.{} was over allocated".format(dev, res)
+                    # TODO(lhc): Due to floating point, it doesn't work.
+                    #assert dres[res] <= dev.resources[res], "{}.{} was over deallocated".format(dev, res)
+                    #assert dres[res] >= 0, "{}.{} was over allocated".format(dev, res)
```
sestephens73, line 1178: Gross. Can we make these integers?

parla/task_runtime.py
nicelhc13: I checked it by printing a thread and it was called only by the scheduler. But let me double check this.

parla/task_runtime.py
nicelhc13: one possible way is use the default vcu sum as the number of threads or something bigger numbers, not 1.
@dialecticDolt  is it fine?

dialecticDolt: Also gross. We might add the resources back up and exceed the original amount bc of rounding errors and the order that tasks complete matters too. 
Probably using: https://docs.python.org/3/library/fractions.html ? might be the cleanest way to handle this.

parla/task_runtime.py
dialecticDolt: It can also be called by any thread that creates a PArray: https://github.com/ut-parla/Parla.py/blob/d215415572eb0b827529e5d142f6dd30b0d6c030/parla/parray/core.py#L52

parla/task_runtime.py
dialecticDolt: No this is intentional. At creation its important to log the persistent memory.

PR #126, opened 2022-05-10
Changing launch mechanism to worker thread callbacks

dialecticDolt:
NOTE: DRAFT PR, DO NOT MERGE. 
Lots of debugging and optimization left to do. Making this PR early so people can edit and comment. 

Also note: This does not fix any bugs in dev, all bugs present in dev should still occur here.  

The scheduling phases have been moved out of the scheduler thread loop into 'worker.run' and the 'spawn' decorator. 
These are 'launch_tasks_callback', 'schedule_tasks_callback', and 'map_tasks_callback'. 
Each has a non-blocking lock such that only 1 worker thread can access each phase at a time. 
Each has some condition that needs to be met for each phase to run. These are currently set to the greediest possible conditions. They will run as long as there might be work available to schedule. 

I have not yet completely killed the old scheduler thread itself. Right now, it just releases the GIL and does nothing. 
When I try to remove it it either:
 (1) makes execution slower or 
 (2) I'm worried about breaking the final runtime synchronization step in scheduler `__exit__`. 

Additionally:
- I have made separate locks for all of the separate counter and queues to prevent contention. 
  (This has significant performance improvements when used with the old scheduler style loop)
- VCUs have been added back in. 

I have only tested on GPU Cholesky 10k, 14k, 20k locally and on Frontera for correctness.
On these tests it is about as fast as the current dev. Although the callback scheduler appears to have higher variability. 

- Synthetic testing is still needed
- CPU testing is still needed. 
- Nested Task testing is still needed

This might currently fail for tasks that are spawned out of dependency order as the first callback will be missed.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -18,7 +18,7 @@ from parla.cpu_impl import cpu
 from parla.dataflow import Dataflow
 
 # Logger configuration (uncomment and adjust level if needed)
-#logging.basicConfig(level = logging.DEBUG)
+# logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 __all__ = ["Task", "SchedulerContext", "DeviceSetRequirements", "OptionsRequirements",
@@ -992,8 +992,16 @@ class WorkerThread(ControllableThread, SchedulerContext):
                             f"[WorkerThread %d] Starting: %s", self.index, self.task.name)
                         self._status = "Running Task {}".format(self.task)
                         self.task.run()
+
+                        # Free self back to worker pool
                         self._remove_task()
                         self.scheduler.append_free_thread(self)
+
+                        # Activate scheduler
+                        self.scheduler.map_tasks_callback()
+                        self.scheduler.schedule_tasks_callback()
+                        self.scheduler.launch_tasks_callback()
+
                     # Thread wakes up without a task (should only happen at end of program)
                     elif not self.task and self._should_run:
                         raise WorkerThreadException(
@@ -1030,6 +1038,7 @@ class ParrayTracker():
 # Two major TODO (ses) items:
 # 1) Fix the mutexes here. There are places where if the GIL switched, stuff could break
 # 2) Provide a way for the mapper to evict if the resourcepool shows a task can't be mapped anywhere and we need to clear space
+# TODO: Add per-device resource locks (from lhc PR). Assume the # of keys and which keys exist does not change from initialization..?
 class ResourcePool:
     # Importing this at the top of the file breaks due to circular dependencies
     from parla.parray.core import CPU_INDEX, PArray
@@ -1101,8 +1110,8 @@ class ResourcePool:
                 dres = self._devices[d]
 
                 # Workaround stupid vcus (I'm getting rid of these at some point)
-                if d.architecture.id == 'gpu' and name == 'vcus':
-                    continue
+                # if d.architecture.id == 'gpu' and name == 'vcus':
+                #    continue
 
                 if amount > dres[name]:
                     is_available = False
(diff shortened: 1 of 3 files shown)


dialecticDolt:
Err, I'm sorry for all the line changes. Apparently my VSCode was set to reformat lines over a certain length on save. 
I'll fix these soon.

dialecticDolt:
Next steps:
- Allow subset of queues to be scheduled instead of looping through whole queue each time

- Tune conditionals for each stage

- Optimize implementation of scheduler.

parla/task_runtime.py
```diff
@@ -1417,12 +1455,12 @@ def __init__(self, environments: Collection[TaskEnvironment], n_threads: Optiona
         # Start the scheduler thread (likely to change later)
         self.start()
 
-    @property
-    @lru_cache(maxsize=1)
+    @ property
```
nicelhc13, line 1458: should be removed right?

parla/task_runtime.py
```diff
@@ -1923,46 +1981,144 @@ def _schedule_tasks(self):
...
+                        if dev.architecture.id == "cpu" or self.get_launched_datamove_task_count(dev) < (self._num_colocatable_tasks + 1):
+                            self._launch_task(datamove_queue, dev)
+
+    def map_tasks_callback(self):
+        """ This is a callback function for the mapper. Called by WorkerThread and Spawn Decorator to trigger scheduler execution"""
+
+        # Check runtime condition
+        # Are there tasks that have not yet been mapped?
+        # Is the ready queue under a certain threshold?
+        ready_queue_threshold = 50
```
nicelhc13, line 2022: not gonna use this?

parla/task_runtime.py
```diff
@@ -1824,7 +1881,7 @@ def _construct_datamove_task(self, target_data, compute_task: ComputeTask, opera
     def _map_tasks(self):
         # The first loop iterates a spawned task queue
         # and constructs a mapped task subgrpah.
-        #logger.debug("[Scheduler] Map Phase")
+        # logger.debug("[Scheduler] Map Phase")
```
nicelhc13, line 1884: the below while loop (line 1829) needs to be break when there is no task on the queue, doesn't it?

nicelhc13:
This looks good to me. One minor concern is why we need to maintain the scheduler thread as a thread. Why don't we make this a just simple class? (Maybe I am missing something that we discussed.) (Now I got it. You just want to avoid the case that breaks the __exit__, right? Now I understand why you said it. But our ideal option is still to make this a general class, not a thread, right?)

parla/task_runtime.py
```diff
@@ -1755,19 +1798,33 @@ def update_launched_task_count(self, task, dev, counts):
...
             self._device_launched_datamove_task_counts[dev] += counts
 
-    def update_mapped_task_count(self, task, dev, counts):
-        with self._monitor:
+    def get_launched_compute_task_count(self, dev):
+        with self._launched_count_monitor[dev]:
+            return self._device_launched_compute_task_counts[dev]
+
+    def get_launched_datamove_task_count(self, dev):
+        with self._launced_count_monitor[dev]:
```
nicelhc13, line 1806: it should be `_launched_count_..` not launced

parla/task_runtime.py
dialecticDolt: oh I'm not sure. I think vscode just added a space here, I'll fix it in the edits.

parla/task_runtime.py
dialecticDolt: Thanks!. Thats what I get for only testing with a manual test case.

parla/task_runtime.py
dialecticDolt: Added it as a potential condition to make the callback run less often. But disabled all of the other checks for debugging. In this commit, all callbacks run as often as they can.

parla/task_runtime.py
dialecticDolt: This is checked at L1888 because `_dequeue_spawned_task` returns None if it hits an IndexError.

nicelhc13 (approved):
LGTM

PR #127, opened 2022-05-17
Add Parla-corresponding Dask Cholesky on GPU and CPU

nicelhc13:


examples/cholesky/dask_gpu_cholesky.py
```diff
@@ -0,0 +1,158 @@
...
+    print("num_blocks:", num_blocks)
+    for i in range(num_blocks):
+        a_ii = (a.name, i, i)
+        if i > 0:
+            prevs = []
+            for k in range(i):
+                prev = name_lt_dot, i, k, i, k
+                dsk[prev] = (syrk, (name, i, k))
+                prevs.append(prev)
+            a_ii = (operator.sub, a_ii, (sum, prevs))
```
dialecticDolt, line 99: I think this sum might cause a CPU copy (mostly leaving this as a note to check this) before I forget.

PR #131, opened 2022-09-07
Fix scheduler early termination bug

nicelhc13:
The current callback-based scheduler terminates earlier before all tasks complete.
Below scenario is the main reason:

1. One thread, t0, starts a launch step. It acquires a launch lock.
2. Other free thread tries to enter the launch step. But it skips that since t0 already got the lock.
3. t0 finishes the step's main loop (but not release the lock yet).
4. At the same time, a running task completes, notifies its dependent tasks, and tasks whose all dependencies are resolved are enqueued onto a device queue.
5. But, t0 releases the lock after it exited the loop, and no threads could enter the mutual section. There is no free thread to start the callbacks, but the enqueued tasks are still remained.

To resolve this problem, we need to integrate all callbacks into a single mutual section. And we need to change non-blocking lock to blocking lock.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -1010,17 +1010,15 @@ class WorkerThread(ControllableThread, SchedulerContext):
                         logger.debug(
                             f"[WorkerThread %d] Starting: %s", self.index, self.task.name)
                         self._status = "Running Task {}".format(self.task)
+                        self.scheduler.incr_running_tasks() 
                         self.task.run()
+                        self.scheduler.decr_running_tasks()
 
                         # Free self back to worker pool
                         self._remove_task()
                         self.scheduler.append_free_thread(self)
-
                         # Activate scheduler
-                        self.scheduler.map_tasks_callback()
-                        self.scheduler.schedule_tasks_callback()
-                        self.scheduler.launch_tasks_callback()
-
+                        self.scheduler.start_scheduler_callbacks()
                     # Thread wakes up without a task (should only happen at end of program)
                     elif not self.task and self._should_run:
                         raise WorkerThreadException(
@@ -1392,6 +1390,9 @@ class Scheduler(ControllableThread, SchedulerContext):
         # Start with one count that is removed when the scheduler is "exited"
         self._active_task_count = 1
 
+        # Track the number of running tasks
+        self._running_task_count = 1
+
         # For load-balancing purposes
         self._active_compute_task_count = 0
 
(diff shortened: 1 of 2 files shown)


dialecticDolt:
This does fix the bug of tasks being left behind by the callback, however, I really don't think we should solve it this way. 
This forces all tasks to wait on the entire callback phase even when scheduled work is available. If we put a more expensive mapping function, it will no longer be hidden by overlap. 

I propose, instead, that at step 4, the completed task checks if it is the last task that will be able to execute the callbacks and if so execute them repeatedly until this is no longer the case.

nicelhc13:
That sounds great but I am concerned that race still could happen between tail_task_check() -> release().

nicelhc13:
Fixed based on your suggestion.

parla/task_runtime.py
```diff
@@ -2115,6 +2129,20 @@ def _launch_tasks(self):
...
 
+    def start_scheduler_callbacks(self):
+        map_succeed = True
+        schedule_succeed = True
+        launch_succeed = True
+        map_succeed = self.map_tasks_callback()
+        schedule_succeed = self.schedule_tasks_callback()
+        launch_succeed = self.launch_tasks_callback()
+        while self.no_running_tasks():
+            if not map_succeed and not schedule_succeed and not launch_succeed:
```
dialecticDolt, line 2140: Commenting to check if this check is necessary/correct? e.g. is L2139 sufficient
- Are there edge cases where we'd want this loop more than once (assuming all phases execute)?

parla/task_runtime.py
```diff
@@ -1010,17 +1010,15 @@ def run(self) -> None:
                         logger.debug(
                             f"[WorkerThread %d] Starting: %s", self.index, self.task.name)
                         self._status = "Running Task {}".format(self.task)
+                        self.scheduler.incr_running_tasks() 
                         self.task.run()
+                        self.scheduler.decr_running_tasks()
 
                         # Free self back to worker pool
                         self._remove_task()
                         self.scheduler.append_free_thread(self)
```
dialecticDolt, line 1019: I think freeing the thread before calling the callbacks is correct. 
I know we talked about it earlier, but I'm pretty sure this is good as then itself is available to be assigned in the launching phase.

parla/task_runtime.py
nicelhc13: That is for avoiding too many unncessary threads to loop over this main loop. I think only thread who blocked other threads needs to iterate.

parla/task_runtime.py
nicelhc13: I see. I think I mispoint to this part for other errors. So the current state is correct, isn't it?

dialecticDolt:
lgtm

PR #132, opened 2022-09-15
Fix scheduler: infinite loop

nicelhc13:
This fixes two bugs:

1. inifite loop:

1-1. one thread keeps iterating scheduler loop.
1-2. assume that # of running tasks is 0.
1-3. only one task is ready to run (all others are dependent on this)
1-4. this task is assigned to the thread iterating the scheduler loop.
1-5. infinite loop.

2. race condition when assigning a task to a thread
assign_task() is called by a thread, but the target thread checks a task without mutex.

### DIFF
--- a/parla/task_runtime.py
+++ b/parla/task_runtime.py
@@ -997,10 +997,10 @@ class WorkerThread(ControllableThread, SchedulerContext):
                     component.initialize_thread()
                 while self._should_run:
                     self._status = "Getting Task"
-                    if not self.task:
+                    with self._monitor:
                         logger.debug(
                             "[%r] Blocking for a task: (%s)", self, self._monitor)
-                        with self._monitor:
+                        if not self.task:
                             self._monitor.wait()
                         logger.debug(
                             "[WorkerThread %d] Waking up.", self.index)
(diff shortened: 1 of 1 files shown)


parla/task_runtime.py
```diff
@@ -2218,7 +2221,7 @@ def launch_tasks_callback(self):
         # Check runtime conditions
         # Are there any tasks to launch?
         # Are there any free worker threads?
-        condition = len(self._free_worker_threads) > 0
+        condition = len(self._free_worker_threads) > 0 and self.num_active_tasks() != 0
```
dialecticDolt, line 2225: Everything looks good. 

This is the only part that confuses me though, if a worker thread has been assigned an active task it shouldn't be free. (e.g its popped off of the queue in _launch_task). Is this a redundant check?

dialecticDolt:
Should be safe to merge 👍

parla/task_runtime.py
nicelhc13: I think that problem has been resolved. A thread appends itself to the free list before starting the scheduler callback. Or is there anything that I am missing? Let me know then I will check it. For now, let me merge this PR.

PR #136, opened 2022-10-21
Update installation and usage instructions

dialecticDolt:
Just updating and restructuring the README on the landing page. (So its better for SC)


dialecticDolt:
This is branched from the tutorial update so I'm adding that back as the upstream for this.

PR #141, opened 2022-12-01
Remove incorrect task mapping flag setting 

nicelhc13:
These flag setting causes an error that enqueues tasks whose data movement tasks are not yet created to ready queues, while dependencies notify them. It is dangerous since data movements are not happened. A solution is to set the flag after data movement tasks are created.


dialecticDolt (approved):
Makes sense. Thanks for catching this!

PR #143, opened 2023-01-06
fix: a bug when using parray slicing within spawn

yinengy:
sometimes device has no copy but still need a parray slices view object in advance.
So the mechanism is modified such that even if there is no copy on the device, it is still possible to slice a parray, which create a view of origin copy
Example:
```
A = parray(numpy_array)
@spawn(inout=[A], placement=gpu(0))
def task1():
    ... a task move A to GPU. free CPU copy ...
@spawn(inout=[A[0]], placement=gpu(1))
def task2():
    error: cpu's copy is freed -> `inout=[A[0]]` trigger an exception,
    since it try a slice A at CPU (spawn() itself is happened at outer CPU task)
```

### DIFF
--- a/parla/parray/memory.py
+++ b/parla/parray/memory.py
@@ -362,8 +362,24 @@ class MultiDeviceBuffer:
             global_slices: slice/ints/tuple/list<int>, use the same format as advance indexing of numpy
 
         Return
-            :class:`cupy.ndarray` or :class:`numpy.array` object
+            :class:`cupy.ndarray` or :class:`numpy.array` object or `None` if there is no copy at that device
         """
+        # check if a copy exists at this device
+        #
+        # This is needed for the usage of @spawn() annotation
+        # sometimes device has no copy but still need a parray slices view object in advance
+        # Example:
+        # A = parray(numpy_array)
+        # @spawn(inout=[A], placement=gpu(0))
+        # def task1():
+        #   ... a task move A to GPU. free CPU copy ...
+        # @spawn(inout=[A[0]], placement=gpu(1))
+        # def task2():
+        #   error: cpu's copy is freed -> `inout=[A[0]]` trigger an exception,
+        #   since it try a slice A at CPU (spawn() itself is happened at outer CPU task)
+        if self._buffer[device_id] is None:
+            return None
+
         # check if there is a mapping
         if self._indices_map[device_id] is None:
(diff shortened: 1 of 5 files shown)


nicelhc13:
question; in your example, you mention "free CPU copy". Is it deterministic or just an assumption?

parla/parray/memory.py
```diff
@@ -362,8 +362,24 @@ def get_by_global_slices(self, device_id: int, global_slices: SlicesType):
...
+        # check if a copy exists at this device
+        #
+        # This is needed for the usage of @spawn() annotation
+        # sometimes device has no copy but still need a parray slices view object in advance
+        # Example:
+        # A = parray(numpy_array)
+        # @spawn(inout=[A], placement=gpu(0))
+        # def task1():
+        #   ... a task move A to GPU. free CPU copy ...
+        # @spawn(inout=[A[0]], placement=gpu(1))
```
nicelhc13, line 376: then A[0] is from gpu0? (IIUC, this originally was from CPU and since the CPU instance was freed it should be from GPU. Is it true? or am I missing?)

parla/parray/memory.py
```diff
@@ -362,8 +362,24 @@ def get_by_global_slices(self, device_id: int, global_slices: SlicesType):
...
+        # def task2():
+        #   error: cpu's copy is freed -> `inout=[A[0]]` trigger an exception,
+        #   since it try a slice A at CPU (spawn() itself is happened at outer CPU task)
+        if self._buffer[device_id] is None:
+            return None
```
nicelhc13, line 381: or is it just returning None? does it mean that the dependency related to A[0] was removed?? In the above example, what task is the task2 dependent?

yinengy:
> 

it is deterministic on some application when GPU get exclusive access to a parray and owner becomes GPU, then cpu copy is evicted and freed

parla/parray/memory.py
yinengy: Yes you are right

parla/parray/memory.py
yinengy: This is a helper function. And when it return None, the caller will process the result and know this device has no copy at all, so it could generate a view object without accessing the buffer. See parray.__get_item__

nicelhc13 (approved):
LGTM

